PR #11, opened 2014-08-14
Enum variants should be read in as an array

erickt:
This patch adds support for serializing/deserializing variants that have more than one element in them. This is a breaking change, as it serializes the variant arguments as a list, even if it has only one element in it.


alexcrichton:
Closing as I think that enums may just not lend themselves in a lossless way to the TOML configuration format.

PR #43, opened 2015-01-03
Updated serialize dependency

vhbit:
In other case using `toml-rs` causes build of old `rustc-serialize` which is not 
compatible with rust's master


alexcrichton:
Thanks!

PR #44, opened 2015-01-06
Remove needless 'old_orphan_check' feature flag

saneyuki:
With build on `rustc 0.13.0-nightly (ad9e75938 2015-01-05 00:26:28 +0000)`, we needless this feature flag.


saneyuki:
Sorry. This was hasty conclusion....

PR #46, opened 2015-01-07
Update to master

vhbit:
- features
- macro reform
- String/Show split in formatter


alexcrichton:
Thanks! Just waiting for a new nightly now :)

vhbit:
It is actually outdated, just pushed a version which works with today's master

PR #57, opened 2015-03-15
Fix for newest rust nightly

mre:


alexcrichton:
Thanks, but this should basically naturally happen regardless. A specification of `0.3.0` is actually "0.3.0 compatible" as defined by semver, and in this case 0.3.3 is compatible so this crate should automatically use the newer version of rustc-serialize with a `cargo update`. Thanks for the PR though!

PR #61, opened 2015-04-20
Serde

oli-obk:
cc https://github.com/erickt/rust-serde/issues/1

> In general I may just be pretty lost and doing bonkers things in my Deserialize implementation, so any tips would definitely be appreciated!

I hope this is what you meant :) I commented on what I couldn't fix.

The remembering of unknown fields is a little messy, but serde doesn't have nice error-recovery.


alexcrichton:
Awesome, thanks! I'll see if I can't patch up the remaining tests soon.

PR #63, opened 2015-06-04
Disallow table redefinitions

vosen:
Currently, in some cases `toml-rs` allows for redefinitions of tables. Eg. this is currently being accepted:
```
[a]
foo="bar"
[a.b]
foo="bar"
[a]
baz="bar"
```
This PR fixes that.

### DIFF
--- a/src/decoder/rustc_serialize.rs
+++ b/src/decoder/rustc_serialize.rs
@@ -3,7 +3,7 @@ use std::mem;
 
 use super::{Decoder, DecodeError};
 use super::DecodeErrorKind::*;
-use {Value, Table};
+use Value;
 
 impl rustc_serialize::Decoder for Decoder {
     type Error = DecodeError;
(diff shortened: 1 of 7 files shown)


alexcrichton:
Thanks for the PR! I must have missed this when implementing the spec originally :(. Unfortunately the change to `Table` is likely to be quite a large breaking change, so I would rather not change the AST representation of the TOML, but instead only modify the parser. Could the parser keep a side table of "defined tables" and return an error if a table shows up more than once?

vosen:
Hmm, I originally thought about just stripping the flag after parsing, but keeping a side table should be much simpler.
I'll get on it.

vosen:
Turns out building a side table of table-definitions requires copying most of `insert_table` and `insert_array` to handle cases like
```
[[foo]]
  [foo.bar]

[[foo]]
  [foo.bar]
```
So I've ended up converting internal AST (with table-definitions tracking) to user-facing AST, because that required copying less complicated stuff.

alexcrichton:
Thanks!

PR #66, opened 2015-08-07
Implement Deserialize for Value

roguePanda:
Since `Value` implements `Serialize` already, it seems nice to have this, at least for symmetry. I don't think the same thing would be possible with rustc_serialize since its model requires knowing what to read. The implementation is basically a strait port of how Serde handles [deserialization](https://github.com/serde-rs/serde/blob/199ed417bd6afc2071d17759b8c7e0ab8d0ba4cc/serde_json/src/value.rs#L265-L347) for its own JSON type


alexcrichton:
Thanks!

PR #67, opened 2015-08-13
Merge with HEAD and update serde to pass all tests

erickt:
This gets the serde test suite written back in April to pass all tests. There were a couple major things done:

* Port to serde 0.5.
* Switched over to using 0.5's new newtype struct support, so that a type like `struct Foo(u32)` doesn't have a wrapper struct/tuple.
* Added a `Deserialize::visit_enum` that matches the rustc-serialize's pattern of trying to deserialize a tuple into each variant.
* In order to prevent floats from being casted into integers (and vice versa), I added overrides for `Deserialize::visit_{usize,u8,u16,..,i8,i16,...}` to call a new `Deserialize::visit_i64`, and the same for `f32` and `f64`.
* Rewrote the serde `MapVisitor` to report the field when having an error.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,11 @@
 language: rust
+rust:
+  - 1.0.0
+  - beta
+  - nightly
 sudo: false
+before_script:
+  - pip install 'travis-cargo<0.2' --user && export PATH=$HOME/.local/bin:$PATH
 script:
   - cargo build --verbose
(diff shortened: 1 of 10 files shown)


alexcrichton:
Awesome, thanks so much @erickt! I've pushed this all to master and have deleted the `serde` branch of toml-rs!

PR #72, opened 2015-09-25
Implement Value::lookup for empty path to return self

kerhong:
Currently there is no way to lookup `self`. Because of `"".split('.').next() == Some("")`, the current functionality attempts to find children with key `""`.

This implements `Value::lookup` with empty string to return `self`.


alexcrichton:
Thanks!

PR #75, opened 2015-10-07
Allow BOM (Byte order mark) in toml

andersforsgren:
If any valid UTF-8 encoded file should be possible to parse, then the parser must be able to handle a leading Byte Order Mark (BOM).

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -190,6 +190,14 @@ impl<'a> Parser<'a> {
         false
     }
 
+    // Consumes a BOM (Byte Order Mark) if one is next
+    fn bom(&mut self) -> bool {
+        match self.peek(0) {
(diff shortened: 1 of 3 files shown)


alexcrichton:
It looks like the TOML format [may now allow a BOM](https://github.com/toml-lang/toml/issues/42#issuecomment-47171327) character (to simplify this level of parsing), so this may want to be turned off by default (I'm fine with it being an extension though). cc @burntsushi though, still feeling the same about the BOM?

BurntSushi:
@alexcrichton I probably spoke under the misguided impression that a BOM didn't exist for UTF-8. If it does, and the TOML spec says "UTF-8," then it seems quite legal to permit it. :-)

alexcrichton:
Hm right, good point! Seems legit to me, thanks @andersforsgren!

PR #79, opened 2015-12-15
Remove floats from the Decoder.toml field when they get parsed

silvo38:
Currently, floats never get removed from the `toml` field, even when they are parsed successfully. read_f64 was missing a call to toml.take()


alexcrichton:
Thanks!

PR #84, opened 2016-01-21
Address issues found by rust-clippy

frewsxcv:


src/parser.rs
```diff
@@ -709,26 +709,26 @@ impl<'a> Parser<'a> {
         }
         let mut it = date.chars();
         let mut valid = true;
-        valid = valid && it.next().map(is_digit).unwrap_or(false);
```
alexcrichton: I tend to prefer `map(...).unwrap_or(...)` as the ordering of the arguments makes more sense to me, could these (and the above) changes be reverted?

frewsxcv: Addressed in the latest force push

alexcrichton:
Thanks!

PR #89, opened 2016-03-27
Allow advanced lookups in lookup and lookup_mut

BourgondAries:
These commits allow one to search for tables with unconventional names. Example:

```rust
let table: toml::Value = r#"[table."name.other"] value = 0"#.parse().unwrap();
let look = table.lookup(r#"table."name.other".value"#).unwrap();
```

These commits have some breaking changes for the tests, as sending a raw " into lookup will no longer work correctly. One must send `r#""\"""#` or `"\"\\\"\""` instead.

Performance: It may be possible to have a slight performance decrease for all lookups because each entry is now parsed and put into a String. This can be remedied by creating a lookup_simple method that takes simple names only.

Bugs discovered: the parser fails to parse "" and '' as a key name.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -292,22 +292,7 @@ impl<'a> Parser<'a> {
 
     // Parse an array index as a natural number
     fn array_index(&mut self) -> Option<String> {
-        let mut index = String::new();
-        while let Some((_, ch)) = self.peek(0) {
-            match ch {
-                v @ '0' ... '9' => {
-                    if !self.eat(v) {
-                        return None
-                    }
-                    index.push(v);
-                }
-                _ => return Some(index),
-            }
-        }
-        if index.len() > 0 {
-            return Some(index);
-        }
-        None
+        self.integer(0, false, false)
     }
 
     /// Parse a path into a vector of paths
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks! Can you add some negative tests for `lookup` as well for when parsing fails?

src/parser.rs
```diff
@@ -290,6 +290,46 @@ impl<'a> Parser<'a> {
         }
     }
 
+    // Parse an array index as a natural number
+    fn array_index(&mut self) -> Option<String> {
```
alexcrichton: I think there's a method to parse integers somewhere else, perhaps that could be leveraged here?

BourgondAries: I think parsing integers accepts the +xx and -xx notation, which I don't think should be applicable. I'll need to look through the parser methods...

BourgondAries:
Added a bunch of negative tests and some positive ones.

I tried avoiding `#[should_panic]` as these tests easily break down if the toml is parsed wrong.

I'm a little uncertain of what convention to use in the tests.

alexcrichton:
Looks great to me, thanks!

PR #93, opened 2016-04-03
Fix wrong handling of unicode escape code

trlim:
Current parser generates error below if a string contains unicode escape code followed by some unicode character, `"\u03B4α"` for example.

```
expected 4 hex digits after a `u` escape
```

This is an attempt to fix it.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -1,4 +1,3 @@
-use std::ascii::AsciiExt;
 use std::char;
 use std::collections::BTreeMap;
 use std::error::Error;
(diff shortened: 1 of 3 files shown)


alexcrichton:
Oh dear, thanks for the report! I think that this may not quite be the right fix as this may cause the parser to panic in the case of something like `\uαααα`. The `is_ascii` check was used to test whether the following indexing operation would panic or not, but perhaps that can be done via a different method?

trlim:
Oh I should have read `str` reference doc more carefully.

Then how about this? I took it from [StackOverflow](http://stackoverflow.com/questions/31799482/efficiently-extract-prefix-substrings) and slightly modified it.
```
let num = if num.char_indices().nth(len).map(|(i, _)| i).unwrap_or(0) == len {
```

alexcrichton:
Looks good to me!

trlim:
Updated

alexcrichton:
Thanks again!

PR #94, opened 2016-04-17
Require newline after table

zofrex:
Currently this library accepts this TOML as valid:

```
[table] key = "value"
```

I'm pretty sure this is not valid TOML. The [spec](https://github.com/toml-lang/toml/blob/master/versions/en/toml-v0.4.0.md#table) says:

> They appear in square brackets on a line by themselves

And pytoml rejects the sample TOML above as invalid.

This pull request changes the library to reject TOML that is malformed in this way.

I am not particularly attached to this particular solution, but I'd rather present a PR than a bug report, so I gave it my best, incredibly clumsy, shot. In particular I'm pretty sure the line numbers I put into the ParserError are not right.

An important note:

Because this was previously accepted as valid, this will be a **breaking change for Cargo**. In particular, several versions of servo/rust-url will no longer work with Cargo if this PR is accepted and Cargo updates its toml-rs dependency (see servo/rust-url#186). I think this is the right thing to do, but I want to highlight the potential impact to the Rust ecosystem.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -447,7 +447,7 @@ mod tests {
 
     #[test]
     fn lookup_advanced_table() {
-        let value: Value = r#"[table."name.other"] value = "my value""#.parse().unwrap();
+        let value: Value = "[table.\"name.other\"]\nvalue = \"my value\"".parse().unwrap();
         let looked = value.lookup(r#"table."name.other".value"#).unwrap();
         assert_eq!(*looked, Value::String(String::from("my value")));
     }
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -271,6 +271,16 @@ impl<'a> Parser<'a> {
                     values: BTreeMap::new(),
                     defined: true,
                 };
+                self.ws();
+                self.comment();
+                if !self.newline() {
+                    self.errors.push(ParserError {
+                        lo: start,
+                        hi: start,
+                        desc: format!("expected a newline after table definition"),
+                    });
+                    return None
+                }
                 if !self.values(&mut table) { return None }
                 if array {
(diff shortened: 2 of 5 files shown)


alexcrichton:
Thanks for the PR! This is indeed unfortunately a bit worrying, especially for Cargo :(

This would probably be a good time to verify this implementation against the [proposed ABNF grammar](https://github.com/toml-lang/toml/pull/236), as I suspect that would certainly have caught this...

I'm fine merging this, but I think that I'll have to release a new major version of this crate. This breaking change definitely shouldn't be picked up by Cargo as-is (as you've pointed out), but I'd like to merge this into Cargo regardless somehow. It'll probably involve parsing with this change, and if that fails parsing with before-the-change. We can then diff the results and provide a warning for a few cycles.

zofrex:
Thanks for the link to the grammar, I hadn't seen this before! My reading of it (which certainly should not be trusted as authoritative) is that the grammar correctly specifies that a newline is required here.

I like the warning idea, but I do wonder if the extra work to do that is justified (i.e. is this a common issue, how many crates would it affect). Could Crater be used to help figure out the impact?

alexcrichton:
Ok, I'm gonna merge this and add an option which can be enabled on the parser for backwards compatibility, thanks again for the PR @zofrex!

gkoz:
This change rejects the following:

```toml
[table] # comment
```

Is that intentional? It doesn't seem as clearly justified; I believe the grammar PR linked above allows comments everywhere; pytoml doesn't seem to reject it.

alexcrichton:
@gkoz Indeed not intentional! I'll add a fix and publish

PR #98, opened 2016-05-19
Accept empty table

zofrex:
Due to requiring an empty line after a table name, the follow valid (I think, having checked against the grammar) input is now rejected, where it was not before:

```
[table]
```

NB that it is still accepted if followed by a blank line:

```
[table]

```

The Cargo.toml files for the crates "decibel" and "water" don't parse without a fix for this.


alexcrichton:
Thanks!

PR #101, opened 2016-05-29
Clippy run and implement Default for State and Enconder

### LINKED ISSUE #102: lookup*() have incorrect lifetime specifiers
Current signature:     pub fn lookup<'a>(&'a self, path: &'a str) -> Option<&'a Value> 

A better signature:     pub fn lookup<'a, 'b>(&'a self, path: &'b str) -> Option<&'a Value> 

Rationale:
- the path has nothing to do with the result.
- the path has no need to live as long as the Value/self.

The current signature is a problem because:
I can't even create a path to use lookup() because of the lifetime restriction. Example:

    pub fn get_config_string<'a, 'b>(config: &'a Value, key: &'b str) -> Result<&'a str, MyError> {
        let path = format!("init.{}", &key);
        config.lookup(&path) <-- fails

Obviously it is "impossible" to create path with the required lifetime 'a.
(I need to limit lookups to subsets of a Value)

I've created my own branch with the 'better signature' and it works fine.
I can't see any downsides. Can you?

Thanks!

leodasvacas:
Fix a bunch of lints indicated by clippy, most common one was useless `format!`. Most notable change is to implement `Default` for `State` and derive it for `Encoder`. Also closes #102.

### DIFF
--- a/src/encoder/rustc_serialize.rs
+++ b/src/encoder/rustc_serialize.rs
@@ -47,10 +47,10 @@ impl rustc_serialize::Encoder for Encoder {
         self.emit_value(Value::Float(v))
     }
     fn emit_char(&mut self, v: char) -> Result<(), Error> {
-        self.emit_str(&*format!("{}", v))
+        self.emit_str(&v.to_string())
     }
     fn emit_str(&mut self, v: &str) -> Result<(), Error> {
-        self.emit_value(Value::String(format!("{}", v)))
+        self.emit_value(Value::String(v.to_string()))
     }
     fn emit_enum<F>(&mut self, _name: &str, f: F)
         -> Result<(), Error>
@@ -98,7 +98,7 @@ impl rustc_serialize::Encoder for Encoder {
         where F: FnOnce(&mut Encoder) -> Result<(), Error>
     {
         let old = mem::replace(&mut self.state,
-                               State::NextKey(format!("{}", f_name)));
+                               State::NextKey(f_name.to_string()));
         try!(f(self));
         if self.state != State::Start {
             return Err(NoValue)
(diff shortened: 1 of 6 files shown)


src/encoder/rustc_serialize.rs
```diff
@@ -179,11 +179,10 @@ impl rustc_serialize::Encodable for Value {
         where E: rustc_serialize::Encoder
     {
         match *self {
-            Value::String(ref s) => e.emit_str(s),
+            Value::String(ref s) | Value::Datetime(ref s) => e.emit_str(s),
```
alexcrichton: Could this stay as a separate branch? It's a little clearer like that and may change one day as well

src/parser.rs
```diff
@@ -168,7 +168,7 @@ impl<'a> Parser<'a> {
     }
 
     fn next_pos(&self) -> usize {
-        self.cur.clone().next().map(|p| p.0).unwrap_or(self.input.len())
+        self.cur.clone().next().map_or(self.input.len(), |p| p.0)
```
alexcrichton: I personally prefer `map().unwrap_or()` as it's easier to read, could this stay as it was?

src/parser.rs
```diff
@@ -188,8 +188,8 @@ impl<'a> Parser<'a> {
     fn expect(&mut self, ch: char) -> bool {
         if self.eat(ch) { return true }
         let mut it = self.cur.clone();
-        let lo = it.next().map(|p| p.0).unwrap_or(self.input.len());
-        let hi = it.next().map(|p| p.0).unwrap_or(self.input.len());
+        let lo = it.next().map_or(self.input.len(), |p| p.0);
+        let hi = it.next().map_or(self.input.len(), |p| p.0);
```
alexcrichton: (same as above)

leodasvacas:
@alexcrichton Done!

PR #103, opened 2016-05-31
lookup() and lookup_mut() lifetime enhancements.

MarkSwanson:
Rationale:

- The path has nothing to do with the result.
- The path has no need to live as long as the Value/self.
- In some cases it can be hard to actually build a path that meets
  the same lifetime requirements as the Value or String slice result.

I submit to you that lifetime elision will not work with my use cases.
I really need the 'path' lifetime to be completely disassociated from the Value or String slice result.
I tried to provide an example in my original Issue. If my example was confusing or poorly worded please ask me to clean it up and I will happily do so.

Thank you.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -180,7 +180,7 @@ impl Value {
     /// let no_bar = value.lookup("test.bar");
     /// assert_eq!(no_bar.is_none(), true);
     /// ```
-    pub fn lookup<'a, 'b>(&'a self, path: &'b str) -> Option<&'a Value> {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! You should actually be able to remove the lifetime annotations entirely as elision should cover this all now.

MarkSwanson:
Well, that would be fantastic! I thought elision only worked if all lifetimes were the same. I'll test this tonight.

MarkSwanson:
You're right (of course).
Thanks for teaching me about this.
The pull request now uses elision.

alexcrichton:
Thanks!

PR #104, opened 2016-07-14
WIP: adjust for seq/map ser revamp

oli-obk:
just leaving this here for future reference so noone needs to repeat any work

### DIFF
--- a/serde-tests/Cargo.toml
+++ b/serde-tests/Cargo.toml
@@ -9,7 +9,6 @@ serde = "0.7"
 toml = { path = "..", features = ["serde"] }
 
 [build-dependencies]
-syntex = "0.33"
 serde_codegen = "0.7"
 
 [lib]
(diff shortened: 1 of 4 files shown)


alexcrichton:
Thanks! Is this an upcoming change in the next version of serde?

dtolnay:
This is planned for serde 0.8.0. We are still working out the exact Serializer API in https://github.com/serde-rs/serde/pull/437.

The point is to make serialization push-based where the Serialize is in charge and the Serializer is passive. The current API is a confusing mix of push-based for primitives and pull-based for seq/map/struct.

Push-based is better for performance and also makes it possible to serialize seq/map/struct without knowing all of the elements up front - see https://github.com/serde-rs/serde/issues/386.

alexcrichton:
Awesome, thanks for the heads up and the implementation! I'll be sure to merge as soon as 0.8.0 is out

Issue #119, opened 2016-11-07
allow configuration when encoding to strings


vitiral:
The use case I am trying to meet is the ability to configure `toml` to use tripple quotes (`'''`) when useful (such as when there are any newlines.

Here is some example code:
```
extern crate rustc_serialize;
extern crate toml;

use toml::{encode_str};

#[derive(RustcEncodable)]
struct MyStruct { foo: isize, bar: String }

const LONG_TEXT: &'static str = r#"
this is some text
here is some more
yay
"#;

fn main() {
    let my_struct = MyStruct { foo: 4, bar: LONG_TEXT.to_string() };
    println!("{}", encode_str(&my_struct));
}
```

This currently prints:
```
bar = "\nthis is some text\nhere is some more\nyay\n"
foo = 4
```

I would like for it to be able to print:
```
bar = '''
this is some text
here is some more
yay
'''
foo = 4
```

vitiral:
I would think the API for this would be nearly identical to [Encoder::new_pretty()](https://doc.rust-lang.org/rustc-serialize/rustc_serialize/json/struct.Encoder.html#method.new_pretty) from `rustc_serialize::json`

alexcrichton:
Agreed would be nice to have!

vitiral:
I can try to work on this if you are on board. I really need it for  [rst](https://github.com/vitiral/rst)

alexcrichton:
Sure!

vitiral:
I take that back... after spending a few hours trying to understand how to do this I am at a loss.

What I want to do is do something similar to `rustc_serialze/src/json.rs::emit_enum_variant` except I only have to do something special when the type is a `Value::String`.

However, this does not seem to be possible, as the `Value` object doesn't exist, instead it is replaced with a `Table` object (`Encoder.toml`). Even if I were able to do a match statement on the Value, I don't have access to the `Writer` object like `rustc_serialize` does (on that note, I'm not entirely sure how you are writing data out at all)

Other options have also not panned out. For instance, I thought of overriding the `encode` method on `Value` (in `src/rustc_serialize.rs`), however THAT has the issue that I don't have access to the `writer` either.

I'm at an impass, I feel like some kind of refactoring has to happen to get this functionality:
 - use a `Value` instead of a `Table` for `Encoder.toml`
 - attach a `Writer` attribute to `Encoder`

Other suggestions would be helpful.

alexcrichton:
[...]

PR #121, opened 2016-11-23
initial attempt at pretty toml strings

vitiral:
this is my initial working method for making toml strings "prettier."

This method is BAD -- it currently makes a completely separate `Printer` object called `PrettyPrinter` when really only a few lines need to be changed.

The issue I can't get around is that `Printer` has a `Formatter` object attached to it... and I have literally no idea how to get one of those without using the `format!` or related macros and overriding `fmt::Display` (I can't do that because I need to insert the pretty options in somehow!)

Do you know how to get a `Formatter` object so I can use it?

Other tasks:
- move the tests to the `tests` folder

Other comments would be appreciated

### DIFF
--- a/src/display.rs
+++ b/src/display.rs
@@ -1,4 +1,6 @@
+use std::fmt::Write;
 use std::fmt;
+use std::string::{String as StdString};
 
 use Table as TomlTable;
 use Value::{self, String, Integer, Float, Boolean, Datetime, Array, Table};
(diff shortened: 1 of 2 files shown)


vitiral:
closes #119

vitiral:
one possibility would be to attach the formatting options to the `TomlTable` object... not my favorite option, but possible.

src/display.rs
```diff
@@ -207,3 +209,155 @@ mod tests {
...
+
+/// Encodes an encodable value into a "pretty" TOML string.
+///
+/// This function expects the type given to represent a TOML table in some form.
+/// If encoding encounters an error, then this function will fail the task.
+/// 
+/// "pretty" means the following features of the output are changed:
+/// - strings with newlines characters (`\n`) will use the `'''` form
+///     and span multiple lines
+pub fn encode_str_pretty<'a>(tbl: TomlTable) -> Result<StdString, fmt::Error> {
```
alexcrichton: Ideally this would look like:

```rust
impl Value {
    pub fn pretty(&self) -> Pretty {
        // ...
    }
}

pub struct Pretty<'a> {
    inner: &'a Value,
}

impl<'a> fmt::Display for Pretty<'a> {
    // ...
}
```

That way you could just do:

```rust
let s = val.pretty().to_string();
```

or

```rust
let s = format!("toml: {}", val.pretty());
```

alexcrichton: With something like that we could also support builder-style configuration on `Pretty` itself, adding support for various options over time.

alexcrichton:
Closing due to inactivity, but feel free to resubmit!

PR #128, opened 2017-01-05
Added seperate serde functions.

### LINKED ISSUE #117: Provide serialization api like serde_json
Like

```
    let serialized = serde_json::to_string(&point).unwrap();
    let deserialized: Point = serde_json::from_str(&serialized).unwrap();
```

good to have

```
    let serialized = serde_toml::to_string(&point).unwrap();
    let deserialized: Point = serde_toml::from_str(&serialized).unwrap();
```

martinrlilja:
Small change to allow the rustc-serialize and serde feature to be used at the same time (with the helper functions). This introduces four new functions `deserialize`, `deserialize_str`, `serialize` and `serialize_str`, which only takes care of serde. While the old functions now only takes care of rustc-serialize. Therefore, this is a breaking change.

Closes #117

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -83,8 +83,8 @@ pub fn decode<T: ::rustc_serialize::Decodable>(toml: Value) -> Option<T> {
 /// into the type specified. If decoding fails, `None` will be returned. If a
 /// finer-grained error is desired, then it is recommended to use `Decodable`
 /// directly.
-#[cfg(all(not(feature = "rustc-serialize"), feature = "serde"))]
-pub fn decode<T: ::serde::Deserialize>(toml: Value) -> Option<T> {
+#[cfg(feature = "serde")]
+pub fn deserialize<T: ::serde::Deserialize>(toml: Value) -> Option<T> {
     ::serde::Deserialize::deserialize(&mut Decoder::new(toml)).ok()
 }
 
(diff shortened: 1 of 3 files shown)


theduke:
You might as well use the de facto naming conventions for other serde libraries:

Instead of `encode_` / `decode_`: `to_string()` and `from_str()`.

See json_serde for an example.

dtolnay:
:+1: for `to_string` and `from_str`. We are planning to standardize that naming across all Serde formats that deal with bytes: https://github.com/serde-rs/serde/issues/644.

```rust
fn to_writer<W, T: ?Sized>(writer: W, value: &T) -> Result<(), Error>
    where W: Write, T: Serialize;
fn to_string<T: ?Sized>(value: &T) -> Result<String, Error>
    where T: Serialize;
fn to_bytes<T: ?Sized>(value: &T) -> Result<Vec<u8>, Error>
    where T: Serialize;

fn from_reader<R, T>(reader: R) -> Result<T, Error>
    where R: Read, T: Deserialize;
fn from_str<T>(s: &str) -> Result<T, Error>
    where T: Deserialize;
fn from_bytes<T>(v: &[u8]) -> Result<T, Error>
    where T: Deserialize;
```

theduke:
@dtolnay great. 

Most libraries have to_vec() instead of to_bytes right now, though.
But they can just add to_bytes, deprecate to_vec() and drop it for the next major release.

Is there a tracking issue about it?

dtolnay:
The tracking issue is https://github.com/serde-rs/serde/issues/644. Really it's tracking these as trait methods that you implement for the corresponding Serializer and Deserializer, so libraries can keep whatever free functions they had before (`toml::to_vec` or whatever). The point of the traits would be to enable writing code that abstracts over all the possible ways you could convert a type to bytes and back. It doesn't mean we have to deprecate `serde_json::to_string` in favor of `serde_json::Serializer::to_string`.

alexcrichton:
Thanks for the PR! This is a breaking change for those crates which only enable the `serde` feature, so perhaps this could leave the existing functions as-is and just add the new functions?

alexcrichton:
I'd also be totally fine adding the API @dtolnay [mentions](https://github.com/alexcrichton/toml-rs/pull/128#issuecomment-271114608) and then we can just reserve these historical names for rustc-serialize users.

alexcrichton:
I'm going to close this in favor of https://github.com/alexcrichton/toml-rs/pull/137 which I believe has all the pieces here as well, but thanks regardless for the PR!

PR #137, opened 2017-02-08
Rewrite crate with serde support from ground up

alexcrichton:
This commit completely rewrites this crate from the ground up,
supporting serde at the lowest levels as I believe serde support was
intended to do. This is a major change from the previous versions of
this crate, with a summary of changes being:

* Serialization directly to TOML is now supported without going through
  a `Value` first.

* Deserialization directly from TOML is now supported without going
  through a `Value`. Note that due to the TOML format some values still
  are buffered in intermediate memory, but overall this should be at a
  minimum now.

* The API of `Value` was overhauled to match the API of
  `serde_json::Value`. The changes here were to:

  * Add `is_*` accessors
  * Add `get` and `get_mut` for one-field lookups.
  * Implement panicking lookups through `Index`

  The old `index` methods are now gone in favor of `get` and `Index`
  implementations.

* A `Datetime` type has been added to represent a TOML datetime in a
  first-class fashion. Currently this type provides no accessors other
  than a `Display` implementation, but the idea is that this will grow
  support over time for decomposing the date.

* Support for the `rustc-serialize` crate has been dropped, that'll stay
  on the 0.2 and 0.1 release trains.


alexcrichton:
@dtolnay if you feel up for it I'd love a gut check on just a skim of the various implementations here, I'm not sure how much is unidiomatic for serde but at least API-wise I tried to stick as close as I could to serde_json when possible!

src/de.rs
```diff
@@ -0,0 +1,1200 @@
...
+            return Ok(None)
+        }
+
+        let next = self.tables[..self.max]
+                       .iter()
+                       .enumerate()
+                       .skip(self.cur_parent + 1)
+                       .find(|&(_, table)| {
+            table.array && table.header == self.tables[self.cur_parent].header
+        }).map(|p| p.0);
```
dtolnay: This close brace is in a very confusing place. Try something like this:

```rust
let next = self.tables[..self.max]
               .iter()
               .enumerate()
               .skip(self.cur_parent + 1)
               .find(|&(_, table)| {
                   table.array && table.header == self.tables[self.cur_parent].header
               })
               .map(|p| p.0);
```

or rearrange this code to make it clearer.

src/de.rs
```diff
@@ -0,0 +1,1200 @@
...
+                    Some(Token::Keylike(_)) => {}
+                    _ => return Err(self.error(start, ErrorKind::DateInvalid)),
+                }
+            }
+        }
+        let end = self.tokens.current();
+        Ok(&self.tokens.input()[start..end])
+    }
+
+    // TODO: shouldn't buffer up this entire table in memory, it'd be great to
```
dtolnay: Please file a ticket for this and link to it here. That way people reading the code can easily check on the status.

src/de.rs
```diff
@@ -0,0 +1,1200 @@
...
+            self.eat_whitespace()?;
+            if self.eat(Token::RightBrace)? {
+                return Ok(ret)
+            }
+            self.expect(Token::Comma)?;
+            self.eat_whitespace()?;
+        }
+    }
+
+    // TODO: shouldn't buffer up this entire array in memory, it'd be great to
```
dtolnay: Please include a link to a ticket tracking this.

src/lib.rs
```diff
@@ -1,507 +1,169 @@
...
-//! and detailed error messages when things go wrong.
+//! [TOML]: https://github.com/toml-lang/toml
 //!
-//! This implementation currently passes the language agnostic [test suite][2].
+//! This library implements a [TOML] v0.4.0 compatible parser. This crate also
+//! primarily supports the [`serde`] library for encoding/decoding support to
+//! various types in Rust.
 //!
-//! # Example
+//! TOML itself is an simple, ergonomic, and readable configuration format:
```
dtolnay: "a simple"

src/lib.rs
```diff
@@ -1,507 +1,169 @@
...
+//! * `Deserializer for Value`
+//!
+//! This notably means that you can use Serde to deserialize/serialize the
+//! `Value` type as well as the `Datetime` type in this crate. Similarly you can
+//! use the `Deserializer`, `Serializer`, or `Value` type itself to act as
+//! a deserializer/serializer for arbitrary types.
+//!
+//! An example of deserializing with TOML is:
+//!
+//! ```rust,ignore
```
dtolnay: Don't ignore this example. You already have serde_derive as a dev-dependency.

src/lib.rs
```diff
@@ -1,507 +1,169 @@
...
-//! This library also supports using the standard `Encodable` and `Decodable`
-//! traits with TOML values. This library provides the following conversion
-//! capabilities:
+//! Similarly you can serialize types in a similar fashion:
 //!
-//! * `String` => `toml::Value` - via `Parser`
-//! * `toml::Value` => `String` - via `Display`
-//! * `toml::Value` => rust object - via `Decoder`
-//! * rust object => `toml::Value` - via `Encoder`
+//! ```rust,ignore
```
dtolnay: Don't ignore this example.

src/lib.rs
```diff
@@ -1,507 +1,169 @@
...
+//! }
+//!
+//! #[derive(Deserialize)]
+//! struct Keys {
+//!     github: String,
+//!     travis: Option<String>,
+//! }
+//!
+//! fn main() {
+//!     let config = toml::from_str(r#"
```
dtolnay: There is no way this works - I think you need `let config: Config = ...`.

src/ser.rs
```diff
@@ -0,0 +1,1130 @@
...
+        if let State::Table { .. } = self.state {
+            self.dst.push_str("\n");
+        }
+        Ok(())
+    }
+
+    fn emit_key(&mut self, type_: &'static str) -> Result<(), Error> {
+        self.array_type(type_)?;
+        let state = self.state.clone();
+        self.emit_key_(&state)
```
dtolnay: How come we are cloning, taking a reference, and then dropping the clone?

src/ser.rs
```diff
@@ -0,0 +1,1130 @@
...
+    }
+
+    fn emit_key_(&mut self, state: &State) -> Result<(), Error> {
```
dtolnay: Can you rename this function or emit_key to clarify how they are different?

src/ser.rs
```diff
@@ -0,0 +1,1130 @@
...
+                                _variant_index: usize,
+                                _variant: &'static str,
+                                _len: usize)
+                                -> Result<Self::SerializeStructVariant, Self::Error> {
+        Err(Error::DateInvalid)
+    }
+}
+
+#[doc(hidden)]
+pub struct Impossible<T> {
```
dtolnay: The next 120 lines or so could be replaced by [`serde::ser::Impossible<(), Error>`](https://docs.serde.rs/serde/ser/struct.Impossible.html).

src/ser.rs
```diff
@@ -0,0 +1,1130 @@
...
+        match self.nope {}
+    }
+}
+
+struct StringExtractor<'a> {
+    slot: &'a mut String,
+}
+
+impl<'a> ser::Serializer for StringExtractor<'a> {
+    type Ok = ();
```
dtolnay: I think this would be clearer if StringExtractor is a unit struct and StringExtractor::Ok = String.

src/ser.rs
```diff
@@ -0,0 +1,1130 @@
...
+
+    fn serialize_char(self, _v: char) -> Result<(), Self::Error> {
+        Err(Error::KeyNotString)
+    }
+
+    fn serialize_str(self, value: &str) -> Result<(), Self::Error> {
+        if value.contains("\n") {
+            return Err(Error::KeyNewline)
+        }
+        assert_eq!(self.slot.len(), 0);
```
dtolnay: I would recommend `assert!(self.slot.is_empty())` but with the suggestion above we don't need this assertion anymore. `Ok(value.to_owned())`.

src/value.rs
```diff
@@ -0,0 +1,869 @@
...
+    }
+}
+
+impl<I> ops::IndexMut<I> for Value where I: Index {
+    fn index_mut(&mut self, index: I) -> &mut Value {
+        self.get_mut(index).expect("index not found")
+    }
+}
+
+impl From<String> for Value {
```
dtolnay: Let's also provide From\<Datetime>.

src/value.rs
```diff
@@ -0,0 +1,869 @@
...
+    }
+}
+
+impl From<Table> for Value {
+    fn from(val: Table) -> Value {
+        Value::Table(val)
+    }
+}
+
+#[doc(hidden)]
```
dtolnay: I don't like this approach because rustdoc does not give me a precise list of types that I can use as index. I think serde_json's approach is clearer: https://docs.serde.rs/serde_json/value/trait.Index.html

src/value.rs
```diff
@@ -0,0 +1,869 @@
...
+                    Err(de::Error::invalid_length(len, &"fewer elements in map"))
+                }
+            }
+        }
+    }
+
+	fn deserialize_option<V>(self, visitor: V) -> Result<V::Value, ::de::Error>
+		where V: de::Visitor
+	{
+		visitor.visit_some(self)
```
dtolnay: Add a comment here and the other deserialize_option methods that this is because TOML does not have a representation corresponding to null.

dtolnay:
I didn't look at any of the TOML-specific parsing and stuff but I skimmed all the Serde code. Great work! Looks like you got the hang of it.

src/lib.rs
```diff
@@ -1,507 +1,169 @@
...
+//! assert_eq!(value["foo"].as_str(), Some("bar"));
 //! ```
 //!
-//! # Conversions
+//! The `Value` type implements a number of convenience methods and traits,
+//! where the example above is using `FromStr` to parse a `str` into a `Value`.
+//!
+//! ## Deserialization and Serialization
+//!
+//! This crates currently supports [`serde`] 0.9 with a number of
```
shepmaster: "this crate"

src/value.rs
```diff
@@ -0,0 +1,869 @@
...
+        self.as_float().is_some()
+    }
+
+    /// Extracts the boolean value if it is a boolean.
+    pub fn as_bool(&self) -> Option<bool> {
+        match *self { Value::Boolean(b) => Some(b), _ => None }
+    }
+
+    /// Tests whether this value is an boolg
+    pub fn is_boolg(&self) -> bool {
```
shepmaster: What's a `boolg`?

src/ser.rs
alexcrichton: Ideally this'd be `self.emit_key_(&self.state)` but, as `emit_key_` takes a mutable borrow that ends up being a borrowck error (clones to the rescue!)

In this case it's a cheap clone anyway, so not too much to worry about

alexcrichton:
Alright let's do this!

jimmycuadra:
Is there more to do before a crates.io release of 0.3? Eagerly awaiting these changes! (And thank you for all the great work, Alex and David!)

alexcrichton:
Soon! I'd like to take some time today and port Cargo to using this to flesh out any final bugs, and then I'll publish once that test suite is passing. Either that or if the port takes too long I'll publish anyway!

stephank:
I don't think there's a way to get notified of Cargo releases, right? Could you leave us a ping here, @alexcrichton? Much appreciated! :)

For what it's worth, I have [a branch of Portier](https://github.com/portier/portier-broker/pull/139) that is working fine! Granted, we're doing, maybe, basic stuff. :)

alexcrichton:
Certainly will do!

I've got Cargo completely compiling now and am just working through test failures. Once that's all clear though I'll publish.

alexcrichton:
Ok, 0.3.0 is now publish!

Thanks again for all your help @dtolnay!

PR #141, opened 2017-02-09
Minor prose fixes

shepmaster:


src/value.rs
```diff
@@ -113,7 +113,7 @@ impl Value {
         match *self { Value::Boolean(b) => Some(b), _ => None }
     }
 
-    /// Tests whether this value is an boolg
```
alexcrichton: It lives!

alexcrichton:
I have words, the best words

Issue #139, opened 2017-02-09
Fix clippy complaints


dtolnay:
I tried running `cargo clippy` against the serde-upgrade branch and it flags 33 things.

est31:
Fixed by #157

ehuss:
I wouldn't say it's fixed, there are now over 100 warnings.

NickHackman:
In #321 all Clippy Errors are fixed and one Warning remains

epage:
Maintenance of this crate has moved to the https://github.com/toml-rs/toml repo.  As a heads up, we plan to move `toml` to be on top of `toml_edit`, see https://github.com/toml-rs/toml/issues/340.

Closing this out.  If this is still a problem, feel free to recreate this issue in the new repo.

PR #157, opened 2017-03-30
Fix issues indicated by clippy

neosilky:
This also fixes quite a lot of #139.

Let me know if anything isn't preferred style.

### DIFF
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -179,11 +179,10 @@ impl FromStr for Datetime {
                               chars.clone().next() == Some('T') {
             chars.next();
             true
-        } else if full_date.is_none() {
-            true
         } else {
-            false
+            full_date.is_none()
         };
+
         let time = if partial_time {
             let h1 = digit(&mut chars)?;
             let h2 = digit(&mut chars)?;
@@ -299,7 +298,7 @@ impl FromStr for Datetime {
 
 fn digit(chars: &mut str::Chars) -> Result<u8, DatetimeParseError> {
     match chars.next() {
-        Some(c) if '0' <= c && c <= '9' => Ok(c as u8 - '0' as u8),
+        Some(c) if '0' <= c && c <= '9' => Ok(c as u8 - b'0'),
         _ => Err(DatetimeParseError { _private: () }),
     }
 }
--- a/src/de.rs
+++ b/src/de.rs
@@ -687,7 +687,7 @@ impl<'a> Deserializer<'a> {
     }
 
     fn number_or_date(&mut self, s: &'a str) -> Result<Value<'a>, Error> {
-        if s.contains("T") || (s.len() > 1 && s[1..].contains("-")) &&
+        if s.contains('T') || (s.len() > 1 && s[1..].contains('-')) &&
            !s.contains("e-") {
             self.datetime(s, false).map(Value::Datetime)
         } else if self.eat(Token::Colon)? {
@@ -698,7 +698,7 @@ impl<'a> Deserializer<'a> {
     }
 
     fn number(&mut self, s: &'a str) -> Result<Value<'a>, Error> {
-        if s.contains("e") || s.contains("E") {
+        if s.contains('e') || s.contains('E') {
             self.float(s, None).map(Value::Float)
         } else if self.eat(Token::Period)? {
             let at = self.tokens.current();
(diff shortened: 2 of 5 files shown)


src/datetime.rs
```diff
@@ -179,11 +179,8 @@ impl FromStr for Datetime {
                               chars.clone().next() == Some('T') {
             chars.next();
             true
-        } else if full_date.is_none() {
-            true
-        } else {
-            false
-        };
+        } else { full_date.is_none() };
```
alexcrichton: Can this be spread across multiple lines?

src/de.rs
```diff
@@ -1178,10 +1178,8 @@ impl<'a> Header<'a> {
             }
 
             self.tokens.eat_whitespace()?;
-            if self.require_newline_after_table {
-                if !self.tokens.eat_comment()? {
```
alexcrichton: I prefer this construction, can this remain?

src/tokens.rs
```diff
@@ -66,14 +66,13 @@ impl<'a> Tokenizer<'a> {
...
-        return t
+        t
     }
 
     pub fn next(&mut self) -> Result<Option<Token<'a>>, Error> {
         let token = match self.chars.next() {
             Some((_, '\n')) => Newline,
-            Some((start, ' ')) => self.whitespace_token(start),
-            Some((start, '\t')) => self.whitespace_token(start),
+            Some((start, ' ')) | Some((start, '\t')) => self.whitespace_token(start),
```
alexcrichton: I prefer the previous construction, can that remain?

src/tokens.rs
```diff
@@ -101,8 +100,7 @@ impl<'a> Tokenizer<'a> {
     pub fn eat(&mut self, expected: Token<'a>) -> Result<bool, Error> {
         match self.peek()? {
             Some(ref found) if expected == *found => {}
-            Some(_) => return Ok(false),
-            None => return Ok(false),
```
alexcrichton: I prefer this construction, can it remain?

src/tokens.rs
```diff
@@ -214,7 +212,7 @@ impl<'a> Tokenizer<'a> {
     }
 
     pub fn current(&mut self) -> usize {
-        self.chars.clone().next().map(|i| i.0).unwrap_or(self.input.len())
+        self.chars.clone().next().map(|i| i.0).unwrap_or_else(|| self.input.len())
```
alexcrichton: This change is not necessary

src/tokens.rs
```diff
@@ -229,7 +227,7 @@ impl<'a> Tokenizer<'a> {
     }
 
     fn comment_token(&mut self, start: usize) -> Token<'a> {
-        while let Some((_, ch)) = self.chars.clone().next() {
+        for (_, ch) in self.chars.clone() {
```
alexcrichton: This and the one below are semantically different, the current incarnation should remain.

tests/valid.rs
```diff
@@ -47,7 +47,7 @@ fn run(toml: &str, json: &str) {
 
     // Assert toml == json
     let toml_json = to_json(toml.clone());
-    assert!(json == toml_json,
+    assert_eq!(json, toml_json,
```
alexcrichton: This isn't necessary

src/tokens.rs
clarcharr: I filed Manishearth/rust-clippy#1653 for this; there's been a lot of `unwrap_or_else` false positives.

src/tokens.rs
clarcharr: I filed Manishearth/rust-clippy#1654 for this; this is definitely a bug.

tests/valid.rs
clarcharr: `assert_eq` actually provides better error messages, unless I'm mistaken?

clarcharr:
There are actually a few false positives here so I've filed clippy bugs for them.

tests/valid.rs
alexcrichton: Not in this case where the message is specified, no

clarcharr: I was under the impression that `assert_eq!` would show the `Debug` for each operand on failure whereas `assert!` would just show the expression `json == toml_json`.

neosilky:
@alexcrichton Sorry about the late reply... Everything should be fixed :+1:

alexcrichton:
Thanks!

PR #161, opened 2017-04-05
Crate polish

colin-kiegel:
We had a crate polishing workshop today. One of the crates we picked was toml-rs. :-)

Here are some suggestions. If you dislike some changes, I can modify the PR accordingly.


alexcrichton:
Awesome, thanks so much!

PR #165, opened 2017-04-24
Allow to deserialize/serialize into enums

### LINKED ISSUE #164: Deserializing enums
I want to deserialize enum from string, I've seen many people already filed this issue but I am looking for quick solution.

```Rust
#[derive(Deserialize)]
enum Type {
    One,
    Two,
}
```

The error I'm getting:

```Rust
Error {
    inner: ErrorInner {
        kind: Custom,
        line: None,
        col: 0,
        message: "invalid type: string \"One\", expected enum Type",
        key: ["markets"],
    },
}
```

Is it possible to deserialize enum like that? Is there any quick solution to this without rewriting the parser?
I suppose it requires a lot of work if it hasn't been done for such a long time, or maybe it just wasn't worth the time or maybe there is a way to do this without too much effort?

Keats:
Close #164

It currently doesn't work for deserialization and I have no idea where it fails.

```
thread 'parse_enum_string' panicked at 'toml.clone().try_into() failed with invalid type: string "desc", expected enum Sort for key `a`', tests/serde.rs:344
```

Some of the changes are probably not needed/not optimal, I was following the serde docs and serde-json repo trying to make it work first.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -122,8 +122,7 @@ enum ErrorKind {
     /// type.
     Custom,
 
-    /// TODO
-    ExpectedEnum,
+    /// A struct was expected but something else was found
     ExpectedString,
 
     #[doc(hidden)]
@@ -206,16 +205,15 @@ impl<'de, 'b> de::Deserializer<'de> for &'b mut Deserializer<'de> {
     ) -> Result<V::Value, Error>
         where V: de::Visitor<'de>
     {
-        if self.peek_char()? == '"' {
-            // Visit a unit variant.
-            match self.next()?.unwrap() {
-                Token::String { ref val, .. } => {
-                    visitor.visit_enum(val.clone().into_deserializer())
+        if let Some(next) = self.next()? {
+            match next {
+                Token::String { val, .. } => {
+                    visitor.visit_enum(val.into_deserializer())
                 },
                 _ => Err(Error::from_kind(ErrorKind::ExpectedString))
             }
         } else {
-            Err(Error::from_kind(ErrorKind::ExpectedEnum))
+            Err(Error::from_kind(ErrorKind::UnexpectedEof))
         }
     }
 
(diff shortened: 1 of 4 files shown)


Keats:
@dtolnay if you have any idea, I'll take them!

dtolnay:
Thanks! I think let's keep this to supporting unit variants only. If I remember correctly TOML does not want to support the other styles of variants because there are no good conventions for how to represent them in TOML.

Be careful that unit variants are support anywhere that strings are supported. Maybe add a test case that uses unit variants in the keys of a map, for example.

alexcrichton:
Ah yes currently this crate doesn't support deserializing/serializing all types in Rust, only a subset. Enums are one of those that don't have a great representation in TOML, so typically you'd want to implement serialize/deserialize manually and work with strings directly. Inferring strings from bare variants with no arguments though seems fine to me, but I'd be hesitant to start deserializing arbitrary enums

Keats:
I remove the non-unit variant code. It's now failing somewhere in `ValueDeserializer` and I don't know how to deserialize enum there.

Keats:
Comments addressed and tests now passing

alexcrichton:
Thanks!

crackcomm:
Thanks!

Boscop:
Does this mean that enum (de)serialization with this crate works for all kinds of enums now? 
Which kinds of types are still not supported compared to serde_json? Or are all supported now?

alexcrichton:
No currently only C-like enums are supported (where they serialize as the stringified version of their variants) or `#[serde(untagged)]`-style enums. Arbitrary enums still aren't supported.

Boscop:
What about enums like:
```rust
enum Enum {
    Foo,
    Bar(BarCase),
    Baz(BazCase)
}
```

alexcrichton:
No, those are still not supported with purely `#[derive(Deserialize, Serialize)]`

PR #170, opened 2017-05-09
Allow extracting the location of the error

oli-obk:
I'm trying to improve the error reporting in clippy. Since we forward to rustc's error reporting, adding a line + col info to the error would allow rustc to actually hilight the error through the regular mechanisms instead of us producing a "help: some error at file:line" through the `Display` impl of `Error`

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -1064,6 +1064,8 @@ impl<'a> Deserializer<'a> {
 
 impl Error {
     /// Produces a (line, column) pair of the position of the error if available
+    ///
+    /// All indexes are 0-based.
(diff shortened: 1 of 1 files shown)


alexcrichton:
Looks good to me! Can you also document whether line/col are 0 or 1-index based? I think they're both 0, but I forget...

oli-obk:
Done. As stated on [to_linecol](https://github.com/alexcrichton/toml-rs/pull/170/files#diff-bcad83e135b6617746a5472d41ac63e9R1052) they are 0 based

Issue #171, opened 2017-05-10
Add examples


rap2hpoutre:
Some functions do not have an example section:

- https://docs.rs/toml/0.4.0/toml/ser/fn.to_vec.html
- https://docs.rs/toml/0.4.0/toml/ser/fn.to_string.html
- https://docs.rs/toml/0.4.0/toml/de/fn.from_slice.html
- https://docs.rs/toml/0.4.0/toml/de/fn.from_str.html

Though these functions are used in TOML index: https://docs.rs/toml/0.4.0/toml/index.html so it's already documented (but not everywhere).

@alexcrichton Do you think we should (I could submit PR) add example to these functions, or is it overkill/useless?

alexcrichton:
Nah this would be awesome! PRs would be most welcome!

PR #172, opened 2017-05-10
add example for from_str

rap2hpoutre:
First step in order to fix https://github.com/alexcrichton/toml-rs/issues/171: add an example for `toml::de::from_str`

The TOML string used in this example is inspired by the example given on the official TOML repo: https://github.com/toml-lang/toml and code itself is inspired by the example from the index page of this crate: https://docs.rs/toml/0.4.0/toml/index.html#deserialization-and-serialization. I tried to make it a bit shorter and a bit different (in comparison with the index page), maybe I should have only copy/pasted [it](https://docs.rs/toml/0.4.0/toml/index.html#deserialization-and-serialization) (but I think user would prefer different examples across pages). 

The "name" in TOML is an arbitrary choice (not really, in fact it's the name of my daughter, since the original TOML example uses the name of its creator) I can change it if required.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -33,6 +33,37 @@ pub fn from_slice<'de, T>(bytes: &'de [u8]) -> Result<T, Error>
 ///
 /// This function will attempt to interpret `s` as a TOML document and
 /// deserialize `T` from the document.
+///
+/// # Examples
+///
+/// ```no_run
+/// #[macro_use]
+/// extern crate serde_derive;
+/// extern crate toml;
+///
+/// #[derive(Deserialize)]
+/// struct Config {
+///     title: String,
+///     owner: Owner,
+/// }
+///
+/// #[derive(Deserialize)]
+/// struct Owner {
+///     name: String,
+/// }
+///
+/// fn main() {
+///     let config: Config = toml::from_str(r#"
+///         title = 'TOML Example'
+///
+///         [owner]
+///         name = 'Lisa'
+///     "#).unwrap();
+///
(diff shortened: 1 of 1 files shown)


src/de.rs
```diff
@@ -33,6 +33,37 @@ pub fn from_slice<'de, T>(bytes: &'de [u8]) -> Result<T, Error>
 ///
 /// This function will attempt to interpret `s` as a TOML document and
 /// deserialize `T` from the document.
+///
+/// # Examples
+///
+/// ```no_run
```
dtolnay: Do we need `no_run`?

rap2hpoutre: Uh... No, you are right!  I just added it **without thinking** because someone asked me to add it in one other PR: https://github.com/rust-lang/rust/pull/40904/files/5a7e9ad64a160c9ebfb35969f2bdd1065bc154bb#diff-12dfbaa7f22fc4118fe031d0967e78e6

Still I'm not sure I need to remove it or leave it.

rap2hpoutre: In other word: does an example have to be tested?

dtolnay: The reviewer should have explained it better but in that case it was because aborting the test process counts as failing the test. Understandably the test harness does not treat an aborted test as successful. In that case the test was an example for `std::process::abort` so things were different.

Yes examples should always be tested if feasible.

rap2hpoutre: Thanks! Fixed! I understand now!

alexcrichton:
Awesome, thanks!

PR #173, opened 2017-05-10
add example for to_string

### LINKED ISSUE #171: Add examples

rap2hpoutre:
One more step in order to fix #171: add an example for toml::ser::to_string

Once again the TOML example I wrote is inspired by the first example of the TOML readme: https://github.com/toml-lang/toml#example (this time it's about database configuration).


alexcrichton:
Awesome, thanks!

PR #177, opened 2017-05-29
Support unicode in key names

vmchale:
Currently, the following will not parse correctly:

```toml
[features]
default = ["english"]
english = []
français = []
```

However, the following will:

```toml
[features]
default = ["english"]
english = []
francais = []
```

I patched it so that the first is valid, and added in tests to ensure everything went correctly.

Thanks for reviewing and thank you for making this crate!


dtolnay:
[Relevant section from the TOML spec:](https://github.com/toml-lang/toml#keyvalue-pair)

> Keys may be either bare or quoted. **Bare keys** may only contain letters, numbers, underscores, and dashes (`A-Za-z0-9_-`). Note that bare keys are allowed to be composed of only digits, e.g. `1234`, but are always interpreted as strings. **Quoted keys** follow the exact same rules as either basic strings or literal strings and allow you to use a much broader set of key names. Best practice is to use bare keys except when absolutely necessary.
>
> ```toml
> key = "value"
> bare_key = "value"
> bare-key = "value"
> 1234 = "value"
>
> "127.0.0.1" = "value"
> "character encoding" = "value"
> "ʎǝʞ" = "value"
> 'key2' = "value"
> 'quoted "value"' = "value"
> ```

I would prefer not to accept this if it disagrees with the spec.

alexcrichton:
Thanks for the PR! As pointed out by @dtolnay these keys are currently explicitly disallowed by the specification of TOML, so I'm going to close this for now. If the spec changes though I'd be more than willing to merge!

Issue #181, opened 2017-05-30
Deserializing huge numbers into infinity


alanhdu:
`toml` returns a `NumberInvalid` error when to deserialize `inf`, but happily serialize numbers like `99e999` into `inf`:

```rust
extern crate toml;
fn main() {
    let s = "key = 99e9999";
    let data: toml::Value = toml::from_str(s).unwrap();  // fails with "key = inf"
    println!("{}", toml::to_string(&data).unwrap());
}
```

outputs `key = inf`.

Given that `inf` is already explicitly disallowed, I feel like `toml` should also return a `NumberInvalid` error here instead of silently deserializing things into `inf`.

Found with https://github.com/rust-fuzz/targets/pull/72

alexcrichton:
Fixed by https://github.com/alexcrichton/toml-rs/pull/183

PR #183, opened 2017-05-30
Better invalid number handling

alanhdu:
This improves handling of invalid numbers (e.g. `float` or `inf`) by:
- Return an error when deserializing huge numbers like `9e99999` instead of silently deserializing them into `inf`
- Return an error serializing a non-finite number (`nan`, `+inf`, or `-inf`).

Unfortunately, I also ended up adding a variant to the public `toml::se::Error` enum, so this is a breaking change and would need a version bump.

Addresses https://github.com/alexcrichton/toml-rs/issues/181

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -893,6 +893,12 @@ impl<'a> Deserializer<'a> {
         }
         number.parse().map_err(|_e| {
             self.error(start, ErrorKind::NumberInvalid)
+        }).and_then(|n: f64| {
+            if n.is_finite() {
+                Ok(n)
+            } else {
+                Err(self.error(start, ErrorKind::NumberInvalid))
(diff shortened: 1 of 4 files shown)


alanhdu:
So... it's actually pretty unclear to me whether (1) this is the right approach to handling `nan`s and `inf`s with `toml` and (2) if it is the right approach, whether it's worth bumping a major release version number.

@alexcrichton: What are your thoughts?

alexcrichton:
Nah this looks great to me! There's a "nonexhaustive" variant to prevent nonexhaustive matches, so this should be a backwards-compatible approach, and I think it's fine to basically error out on infinite values.

Thanks so much for the fuzzing and looking into this!

PR #184, opened 2017-06-01
Serialize nested array of tables correctly

### LINKED ISSUE #180: Serializing Nested Array of Tables incorrectly
The TOML table equivalent to the following JSON:

```json
{"table": [
  {"inner": [ {} ] }
]}
```

should serialize into

```toml
[[table]]
  [[table.inner]]
```

Instead, it serializes into:
```toml
[[table.inner]]
```

which is equivalent to `{'table': {'inner': [{}]}}`.

Found with https://github.com/rust-fuzz/targets/pull/72.

alanhdu:
When serializing nested table like `[a.b.c]`, `toml` will leave out the super-tables if it can. Unfortunately, if one of the super-tables is actually an array of tables (`[[ a ]]`), then leaving out the super-table changes the value being encoded. For example:

```toml
[[ a ]]
[[ a.b ]]
```

which is equivalent to `{"a": [{"b": [{}]}]}` is currently serialized into

```toml
[[ a.b ]]
```

which is equivalent to `{"a": {"b": [{}]}}`, which is not the same.

To fix this, whenever we emit a table header, we check for  `[[ a ]]` parents and emit table headers for them first if necessary.

Closes #180

### DIFF
--- a/src/ser.rs
+++ b/src/ser.rs
@@ -308,8 +308,28 @@ impl<'a> Serializer<'a> {
             State::Array { .. } => true,
             _ => false,
         };
+
+        // Unlike [..]s, we can't omit [[..]] ancestors, so be sure to emit table
+        // headers for them.
+        let mut p = state;
+        if let State::Array { first, parent, .. } = *state {
+            if first.get() {
+                p = parent;
+            }
+        }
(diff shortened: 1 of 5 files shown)


alexcrichton:
Awesome, thanks so much for both finding and fixing!

PR #187, opened 2017-06-01
More fuzzing fixes

alanhdu:


src/datetime.rs
```diff
@@ -218,7 +218,10 @@ impl FromStr for Datetime {
                     return Err(DatetimeParseError { _private: () })
                 }
                 chars = whole[end..].chars();
-                match format!("0.{}", &whole[..end]).parse() {
+
+                // truncate to picoseconds precision
+                let last = if end > 12 { 12 } else { end };
```
alanhdu: I decided to go with picosecond precision (12 digits), because it's a nice round number and there was already a test that required 10 digits of precision. If you think we need more (or less) precision, I'm happy to tweak this.

src/ser.rs
```diff
@@ -624,9 +621,6 @@ impl<'a, 'b> ser::SerializeMap for SerializeTable<'a, 'b> {
             SerializeTable::Table { ref mut key, .. } => {
                 key.truncate(0);
                 *key = input.serialize(StringExtractor)?;
-                if key.contains('\n') {
-                    return Err(Error::KeyNewline)
-                }
```
alanhdu: Luckily, there's already logic for handling quoting keys when necessary, so all we have to do is to remove this check!

src/ser.rs
```diff
@@ -107,9 +107,6 @@ pub enum Error {
     /// attempted where the key of a map was not a string.
     KeyNotString,
 
-    /// Keys in maps are not allowed to have newlines.
-    KeyNewline,
```
alexcrichton: Ah unfortunately I think *this* is a breaking change (removing an existing variant), but I think it's ok to basically leave a "dead variant" here. Maybe `#[doc(hidden)]` it?

alexcrichton:
Looks great to me! I can't thank you enough for doing all the fuzzing *and* fixing here!

alanhdu:
So on further thought, I think it might make more sense to only store down to the nanosecond level -- both Go and Python only give you nanosecond resolution, and I can't really think of many applications where you'd need picoseconds.

I also took the opportunity to store nanoseconds as a `u32` instead of a `f64`, which I think reads much more naturally. I'm happy to back that commit out if there are problems though!

alexcrichton:
sounds perfect to me, thanks!

PR #190, opened 2017-06-18
Add exhaustive From<T> impls for Value

SergioBenitez:
As discussed on IRC.

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -254,26 +254,24 @@ impl<S: Into<String> + Hash + Eq, V: Into<Value>> From<HashMap<S, V>> for Value 
 }
 
 macro_rules! impl_into_value {
-    ($variant:ident : $T:ty) => (impl_into_value!($variant: $T,););
-    ($variant:ident : $T:ty, $($extra:tt)*) => (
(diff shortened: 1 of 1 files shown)


src/value.rs
```diff
@@ -219,47 +220,62 @@ impl<I> ops::IndexMut<I> for Value where I: Index {
...
+            fn from(val: $T) -> Value {
+                Value::$variant(val $($extra)*)
+            }
+        }
+    )
+}
+
+impl_into_value!(String: String);
+impl_into_value!(Integer: i64);
+impl_into_value!(Integer: isize, as i64);
```
alexcrichton: Could this use `.into()` instead of `as $t` perhaps? That way we can ensure the standard library guarantees lossless conversions.

alexcrichton:
Thanks!

src/value.rs
SergioBenitez: That works for everything _but_ `isize`. Should we remove the `isize` conversion?

alexcrichton: Yeah seems ok to me!

SergioBenitez: Great! Took care of it.

PR #192, opened 2017-06-23
updated toml version in the docs to 0.4.2

debris:
Because of the docs, I used `0.2.1` instead of `0.4.2` in my project.

https://docs.rs/toml/0.4.2/toml/


alexcrichton:
Thanks!

PR #197, opened 2017-07-09
close #119: allow "pretty" configuration

### LINKED ISSUE #119: allow configuration when encoding to strings

vitiral:
This is an MVP of the "pretty" configuration impelmentation. There are some things I would still like to do (like avoid having to clone `Settings`) and there are still a bunch of tests that need to be added but this seems to work 🎉 🎉 

This is necessary for:
- vitiral/artifact#87
- vitiral/artifact#152

I would particularily like feedback from @alexcrichton on:
- Is it okay to have a `Settings` struct instead of a `Formatter` trait, as suggested by @dtolnay. @dtolnay what do you think about this? I see the trait as probably "higher speed" but I don't think that's as much a requirement for toml as it is for json.
- What should the name be? Is `toml::ser::Settings` good?

Improvements needed:
- The `Serializer::pretty` is [mimicking serde_json's][1] implementation, but we should make `pretty` accept a `Settings` struct
- We should add a `to_string_pretty(dst: &mut W, settings: Settings)` method

[1]: https://docs.serde.rs/serde_json/struct.Serializer.html#method.pretty

### DIFF
--- a/src/ser.rs
+++ b/src/ser.rs
@@ -137,18 +137,27 @@ pub enum Error {
     __Nonexhaustive,
 }
 
-/// If given in `Settings` will result in a "pretty array" with the given settings
-/// TODO: add better docs
 #[derive(Debug, Default, Clone)]
-pub struct ArraySettings {
-    indent: u64,
+#[doc(hidden)]
+/// Internal place for holding array setings
+struct ArraySettings {
+    indent: usize,
     trailing_comma: bool,
 }
 
-/// Formatting Settings
-/// TODO: add better docs
+impl ArraySettings {
+    fn pretty() -> ArraySettings {
+        ArraySettings {
+            indent: 4,
+            trailing_comma: true,
+        }
+    }
+}
+
 #[derive(Debug, Default, Clone)]
-pub struct Settings {
+#[doc(hidden)]
+/// Internal struct for holding serialization settings
+struct Settings {
     array: Option<ArraySettings>,
     pretty_string: bool,
 }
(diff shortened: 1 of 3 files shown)


vitiral:
link to #119

alexcrichton:
Thanks for the PR! The approach here seems reasonable to me, but it looks like there's no way to change the formatting settings? Could these perhaps be builders to tweak all the internal settings? (I think `ArraySettings` can be hidden internally as well, right?)

vitiral:
yes, I intentionally didn't include that yet, as I said here:

> but we should make pretty accept a `Settings` struct

Because I didn't know what kind of API you would prefer!

I could definitely do a builder pattern. What do you think about the following API:
```
let settings = toml::ser::Settings::new()
    .pretty_string()
    .array_indent(4)
    .array_trailing_comma(true)
```

The builder is nice because you can add onto it in a backwards compatible way.

Another thing I'd like you to consider is hiding this behind a feature flag for now (`features=beta` or maybe `pretty`) until it gets more use in the wild.

vitiral:
Also, for avoiding cloning settings I was thinking of just using an `Rc` -- any objections?

alexcrichton:
I think actually how about something like:

* `Serializer::new(dst)` stays the same
* `Serializer::pretty(dst)` will serialize a "pretty string" using default pretty
* Otherwise `Serializer` itself has a builder-style pattern for configuring various serialization settings.

How's that sound?

vitiral:
I've added the builder pattern. Let me know if that looks reasonable.

I'd like to look into adding more tests next. Let me know if you have any advice. I think I'd like to focus on:
- running all `invalid` with `pretty()` and make sure they are still invalid.
- focus on `raw-string`, `raw-multiline-string` and some of the `array-` tests and have "pretty" correlaries. I'll keep the json the same but have a "pretty" file alongside the other. So there will be `raw-string.json`, `raw-string.toml` AND `raw-string-pretty.toml`. Let me know if that sounds good.

src/ser.rs
```diff
@@ -194,9 +220,76 @@ impl<'a> Serializer<'a> {
...
+            dst: dst,
+            state: State::End,
+            settings: Settings {
+                array: Some(ArraySettings::pretty()),
+                pretty_string: true,
+            },
+        }
+    }
+
+    /// Enable or Disable pretty strings
```
alexcrichton: Mind expanding this comment and those below to explain what "pretty" means in the context of arrays/strings?

vitiral:
I'm pushing the documentation fixes now. I also added `to_string_pretty` and turned off escaping for `"` characters.

I thought of one possible bug, which is where a user has `'''` somewhere in their string. What should we do with that? Should I replace such instances with `\'\'\'` ... or maybe `''\'`?

vitiral:
Also, I looked over the current tests as they are -- I don't think I can leverage them as they are because *you do not assert that their format doesn't change*. Therefore I'm going to write a few more manual tests to validate everything.

vitiral:
I'm not sure how to fix the failing doctest. I already included `no_run` in the code block, but rust `1.15` is still failing...

alexcrichton:
I think you can use `ignore` on those code blocks? (it looks like TOML is being interpreted as Rust)

alexcrichton:
Thanks!

PR #199, opened 2017-07-22
demonstrate failing test for 'empty' values

vitiral:
This is NOT intended to be merged yet, but rather point add a unit test for something that fails when it should not.


vitiral:
@alexcrichton I tried to dig into this and have to admit I'm pretty lost with the serde implementation. Any guesses on your part about why this is happening?

alexcrichton:
Table headers are currently [lazily emitted](https://github.com/alexcrichton/toml-rs/blob/f6673dfdb3f93bacfbf0c0cab4d78bc4154eec0c/src/ser.rs#L232-L235), which means that if no values are emitted the table header is itself omitted

vitiral:
Can it be changed to fix this bug? Artifact relies on empty objects still
existing.

This worked in toml 0.2

alexcrichton:
Yeah seems fine to me to fix! Just wanted to point out where it was located.

Although... now that I [look closer](https://github.com/alexcrichton/toml-rs/blob/f6673dfdb3f93bacfbf0c0cab4d78bc4154eec0c/src/ser.rs#L668-L671) this should be implemented... Odd! I wonder if it's a top-level vs nested distinction that shouldn't be there?

vitiral:
OK, sounds good. I think I can probably take a shot at this

vitiral:
So I found one of the problems: if a Struct doesn't serialize anything it needed to emit `Error::UnsupportedNone`

Now I'm hitting anoter bug, in `emit_table_header` the `state=State::End`... it seems like this is the default first state? I'm not sure what to do about this.

alexcrichton:
Sorry I don't have time right now to help investigate :(

vitiral:
@alexcrichton no problem :)

I think I've got a possible fix. The problem had nothing to do with `fn end` -- that was the wrong place to handle this case. The issue is when you have an empty struct you need to still emit the header.

> Note: my fix is currently just a proof of concept and slightly hacky... I'll find a better way to insert the header soon hopefully.

However, to get this to work I made empty structs return an `UnsupportedNone` type -- which breaks a unit test.

However, I would argue the unit test was invalid. I inserted a `println!("toml repr:\n{}", toml::to_string(&Foo { a: None }).unwrap());` into `empty_arrays2` test and got `'toml repr:\n'`... in other words nothing was actually serialized!

I would argue this result proves that the "empty struct" has ALWAYS been equivalent to `None` and that this is therefore NOT a breaking change -- it is a bug fix.

vitiral:
Actually, the unit test might be right -- maybe empty types should just return empty strings? Is serializing `None` or `MyStruct { a: None }` as a empty string bad? I don't think so...

vitiral:
I applied the same logic to SerializeMap and hit several bugs. Going to look into them. I think I am onto something with allowing `UnsupportedNone` and treating it correctly.

vitiral:
well... I was completely on the wrong track and you were completely right. It WAS pretty much just the line of code you pointed to -- it wasn't implemented for `SerializeStruct`.

I just added it to `SerializeStruct` and also added a few good tests.

Thanks for the help. The good news is that after going through 5 rabbit holes on this issue I think I really understand what this library is doing and why.

vitiral:
Also, I want you to know that (using this branch) all tests pass for [artifact](https://github.com/vitiral/artifact) as well! Artifact is officially off of rustc serialize!

alexcrichton:
> Thanks for the help. The good news is that after going through 5 rabbit holes on this issue I think I really understand what this library is doing and why.

Oh dear sorry about that! The fix here looks great, though, thanks so much for this!

vitiral:
no problem :)

Any idea when this (and pretty-fmt) will be published? I can just use the master branch until they are but would like to know.

alexcrichton:
Oh I can go ahead and go publish it right now...

vitiral:
oh, that would be great. Thanks!

alexcrichton:
0.4.3 [should be up now](https://crates.io/crates/toml)

PR #201, opened 2017-07-27
demonstrate tripple quote problem

vitiral:
This is to demonstrate the tripple quote problem I theorized: that if there are `'''` in a pretty string there will be issues. However, the issue I hit was not as expected.

The written test prints the following:
```
EXPECTED:                                              
[example]                                            
text = '''                                          
this is the first line                                                                                   
This has a ''\' in it for no reason                                                                      
this is the third line                               
'''                                                 


RESULT:                                                                                                  
[example]                                           
text = '''                                                                                               
this is the first line                              
This has a ''\\' in it for no reason                
this is the third line                              
''' 
```

It seems that the round trip is doubling the number of `\` it finds?

### DIFF
--- a/src/ser.rs
+++ b/src/ser.rs
@@ -450,14 +450,18 @@ impl<'a> Serializer<'a> {
         if ok {
             drop(write!(self.dst, "{}", key));
         } else {
-            self.emit_str(key)?;
+            self.emit_str(key, true)?;
         }
         Ok(())
     }
 
-    fn emit_str(&mut self, value: &str) -> Result<(), Error> {
-        let do_pretty = if self.settings.pretty_string {
-            value.contains('\n')
+    fn emit_str(&mut self, value: &str, is_key: bool) -> Result<(), Error> {
+        let do_pretty = if !is_key && self.settings.pretty_string {
+            // do pretty only if the block contains at least one newline
+            value.contains('\n') 
+                // but not if it contains any of these, as they are not
+                // representable
+                && !value.contains("'''")
         } else {
             false
         };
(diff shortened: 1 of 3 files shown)


src/ser.rs
```diff
@@ -466,7 +466,21 @@ impl<'a> Serializer<'a> {
         } else {
             drop(write!(self.dst, "\""));
         }
+        //let mut quotes: u8 = 0;
```
vitiral: I wrote this code because it was how I was expecting to solve this... it didn't work and instead I got the problem I posted about

vitiral:
interesting... it seems as though the `'''` syntax is actually "raw syntax" and there is NO escapes allowed. I'm pushing the fix now...

vitiral:
ah, acording to the [spec](https://github.com/toml-lang/toml#user-content-spec) this is "multi-line literal strings" and "there is no escaping whatsoever".

I'm wondering if maybe we made a mistake making these the default "pretty" string and should have instead used `"""` (which DOES allow escaping). I actually didn't even realize that `"""` was functional...

Regardless we can include an option. In many ways `'''` is the simpler default since "what you see is what you get" and the need to write `'''` inside your text blob is pretty rare for most formats...

vitiral:
if we continue to use `'''` there are other cases where we should escape things differently... as espcaes simply don't exist!

vitiral:
I think something that COULD be done is we could do round-trip tests with pretty leveraging the existing test infrastructure. That would probably have caught this.

Until then I highly recommend doing the feature flag suggested in #202

vitiral:
ok, I added pretty to valid.rs... I can probably fix these bugs.

vitiral:
Ok, I have done some pretty significant rework and I think we are in a good state:

- I have changed "pretty" to mean:
    - always try to use literal strings. Fall back on `""` if you cant (i.e. control characters or stuff below)
    - if there is a newline, use `'''`
    - if there is not a newline use `'single line`'
    - if there is a single line with `'` in it use `'''single line with ' in it'''`
    - if none of these work, fall back on `""`

This also found several bugs (specifically control characters that were not representable).

I have also changed single item arrays to stay on one line:
```
foo = ['bar']
```
instead of:
```
foo = [
    'bar',
]
```

vitiral:
Having all this logic requried a redesign of `emit_str` for the pretty case. Specifically we now do NOT do `value.contains(...)` calls (which were not very performant, and would be impossibly unperformant for all the checks we now have to do). Instead we reserve a `String` and append onto it one character at a time. If we hit a problem we just return `None` and it gets represented as `"""`.

vitiral:
I'm almost done... don't merge yet

vitiral:
Ok, I think I'm done now... whew

vitiral:
check out the `pretty_tricky` and `pretty_std` for some nice demos

alexcrichton:
Looks great, thanks!

Would it be useful to get this on crates.io as well?

vitiral:
thanks!

I think it would be good to get up soon. With it tied into all those tests I have a lot more confidence in it.

Do you agree with the new rules? I think having pretty mean "use literal as much as possible, use multiline whenever there are newlines" is the way to go.

I think merge this. I would still like to get indentation fixed for embedded arrays and (eventually) arrays of tables -- but that isn't super critical.

alexcrichton:
Yeah the logic sounds great to me! I'll publish a new version.

vitiral:
Thanks! Do you agree with the new rules? I think having pretty mean "use literal as much as possible, use multiline whenever there are newlines" is the way to go.

I think it would be good to get a new point release up ASAP since the old one has bugs and is a different style. This version has way more testing and is better thought out (I actually didn't know that `'` meant "literal" before...).

I would still like to get indentation fixed for embedded arrays and (eventually) arrays of tables -- but that isn't super critical.

alexcrichton:
Oh right yeah sorry meant to clarify, the new rules sound great! A 0.4.4 release should now be published.

Issue #202, opened 2017-07-27
use """ instead of '''


vitiral:
as pointed out in #201 I probably made a mistake in using `'''` instead of `"""`.

We should discuss. I think a bunch of bugs probably were introduced by `'''` and `'''` isn't well tested.

Recomendation: hide `to_string_pretty` and `Serializer::pretty` behind a feature flag and push a new version with that feature off by default ASAP.

I'm sorry that I just asked you to push a new version, I feel really dumb :(

vitiral:
I think this is not the way to go. Closing

vitiral:
I think this is not the way to go. Closing

vitiral:
whoops

PR #203, opened 2017-07-27
related to #202: move 'pretty' behind a beta feature flag

vitiral:
As suggested in #202 we should implement a `beta` feature flag until pretty formatting is more stabilized. We need to decide between `"""` and `'''` and if we choose `'''` (the current) we need several tests and bufixes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ before_script:
   - pip install 'travis-cargo<0.2' --user && export PATH=$HOME/.local/bin:$PATH
 script:
   - cargo test
+  - cargo test --features beta
   - rustdoc --test README.md -L target
   - cargo doc --no-deps
 after_success:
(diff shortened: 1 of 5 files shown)


vitiral:
I feel really bad about this. @alexcrichton please merge this and push a new version so that I can break as few people as possible :(

Another possiblility would be to decide that we DO want to use `'''`, but we want to fix as many bugs as possible as quickly as possible. I can get working on that.

alexcrichton:
Thanks for this! I'm not super worried about the breakage here though in theory because it's just the "pretty" output, right? In that sense I wonder if we could just fix the bug an d issue a new point release?

vitiral:
I think I've got the important bugs fixed. Sorry for freaking out a bit :)

Also, I think `'''` is correct. In fact, I have extended it in code I am about to push to ALWAYS use string literals when possible. Stay tuned.

alexcrichton:
Ok no worries, thanks for your help on this!

PR #204, opened 2017-07-28
reference count settings to bump performance a bit

vitiral:
definitely not an emergency, I just felt like it needed to be done eventually.

Also fixed a couple of docs that I missed.

### DIFF
--- a/src/ser.rs
+++ b/src/ser.rs
@@ -273,15 +273,15 @@ impl<'a> Serializer<'a> {
     ///
     /// Instead of:
     ///
-    /// ```toml,ignore
+    /// ```ignore
     /// single = "no newlines"
     /// text = "\nfoo\nbar\n"
     /// ```
     ///
     /// You will have:
     ///
-    /// ```toml,ignore
-    /// single = 'no newlines'
+    /// ```ignore
+    /// single = "no newlines"
     /// text = '''
     /// foo
     /// bar
@@ -311,13 +311,13 @@ impl<'a> Serializer<'a> {
     ///
     /// Instead of:
     ///
-    /// ```toml,ignore
+    /// ```ignore
     /// array = ["foo", "bar"]
     /// ```
     ///
     /// You will have:
     ///
-    /// ```toml,ignore
+    /// ```ignore
     /// array = [
     ///     "foo",
     ///     "bar",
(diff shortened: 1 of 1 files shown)


alexcrichton:
Oh hm do we need to do this actually? I figured we had enough lifetimes we could just borrow everything?

src/ser.rs
```diff
@@ -234,7 +241,7 @@ impl<'a> Serializer<'a> {
         Serializer {
             dst: dst,
             state: State::End,
-            settings: Settings::default(),
+            settings: Rc::new(Settings::default()),
```
vitiral: `Settings` cannot be a `&` because it has to be returned here (unless we wanted the user to create and pass in the settings... which was the API initially)

vitiral:
Later, new `Serializer` objects are created in `SerializeSeq`, which means that `Settings` has to be cloned.

We could change the pretty API so the `Settings` object is a separate builder that one passes into a `settings` method or something. Then it could be a reference.

vitiral:
So, the `rustdoc` command is failing in rust 1.15 because it apparently won't recognize that the code block is `toml` not `rust`... is it really necessary that rustdoc is backwards compatible?

https://travis-ci.org/alexcrichton/toml-rs/jobs/258811878

alexcrichton:
I wonder could we have something like an internal split? Perhaps something like:

```rust
pub struct Serializer<'a> {
    // what it is today
}

struct BorrowedSerializer<'a, 'b> {
    settings: &'b Settings,
    // same fields as `Serializer`
}
```

and then `Serializer` would just delegate everything to `BorrowedSerializer`?

vitiral:
that sounds doable, I'll see what I can do

vitiral:
By "internal split" I assume you mean that it is not exposed to the user which would work fine if our public API was just `to_string`, `to_string_pretty`, etc.

However, that is NOT our public API. How would you do something like this:
```
    {
        let mut serializer = toml::Serializer::pretty(&mut result);
        serializer.pretty_array_indent(2);
        value.serialize(&mut serializer).unwrap();
    }
```

Are you thinking that `BorrowedSerializer` should be implemented as `Serializer` is now, and then ALL `Serializer` methods just convert to `BorrowedSerializer` and call them?

I guess I could probably do a macro to make this not TOO painful. That sounds pretty boiler-platy to me, just want to make sure that's what you had in mind.

alexcrichton:
Hm yeah I guess if it requires that much boilerplate it's probably not worth it, I was hoping I guess that it wouldn't require so much!

alexcrichton:
Ok, let's merge now and if necessary we can work with borrows later

PR #205, opened 2017-07-30
doc fixes

vitiral:
this fixes the documentation but unfortunately won't work for rustdoc in rust 1.15. We need to decide what to do about that...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ before_script:
 script:
   - cargo test
   - rustdoc --test README.md -L target
-  - cargo doc --no-deps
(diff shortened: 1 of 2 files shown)


alexcrichton:
I'm ok if `cargo doc` doesn't work on 1.15, but we'll need to get CI green here somehow to land

alexcrichton:
Hm that may not be quite the right shell invocation, Travis is still red :(

Maybe a custom definition of `script` on the include of rust 1.15.0?

vitiral:
Sorry, I forgot the `|| echo "skipping cargo doc"`

alexcrichton:
👍

PR #207, opened 2017-08-13
add spaces between array items and test for them

vitiral:
@killercup noticed in killercup/cargo-edit#163

> This format is pretty (😄), except for the missing newlines around these sections

"these sections" being array items. I actually think that this was probably broken before my changes (but maybe not). Never-the-less this PR fixes the issue and adds spaces around array-table items.

Before:
```
[[array]]
key = "foo"
[[array]]
key = "bar"

[table]
key = "baz"
```

After:
```
[[array]]
key = "foo"

[[array]]
key = "bar"

[table]
key = "baz"
```

### DIFF
--- a/tests/pretty.rs
+++ b/tests/pretty.rs
@@ -212,3 +212,51 @@ fn pretty_tricky() {
     println!("\nRESULT:\n{}", result);
     assert_eq!(toml, &result);
 }
+
+const PRETTY_TABLE_ARRAY: &'static str = r##"[[array]]
+key = 'foo'
+
+[[array]]
+key = 'bar'
+
+[abc]
+doc = 'this is a table'
+
+[example]
+single = 'this is a single line string'
+"##;
+
+#[test]
+fn pretty_table_array() {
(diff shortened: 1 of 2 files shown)


tests/pretty.rs
```diff
@@ -212,3 +212,51 @@ fn pretty_tricky() {
...
+    let toml = PRETTY_TABLE_ARRAY;
+    let value: toml::Value = toml::from_str(toml).unwrap();
+    let mut result = String::with_capacity(128);
+    value.serialize(&mut toml::Serializer::pretty(&mut result)).unwrap();
+    println!("EXPECTED:\n{}", toml);
+    println!("\nRESULT:\n{}", result);
+    assert_eq!(toml, &result);
+}
+
+const TABLE_ARRAY: &'static str = r##"[[array]]
```
vitiral: this is technically not a pretty test... I realized a little way through it that the issue had nothing to do with pretty/not-pretty. I could probably move this test to somewhere more relevant

alexcrichton:
Awesome, thanks!

PR #210, opened 2017-08-13
close #209: remove FIXME, appears to have been invalid

### LINKED ISSUE #209: deserialization counts first newline for specific """ block
the `pretty_tricky_non_literal` test found a bug where when the `glass` field from `pretty_tricky` is added it doesn't quite work. I believe the issue is with the deserializer but it needs more investigation.

See the FIXME in `tests/pretty.rs`

vitiral:
I must have been mistaken before, it works fine (weird...)

### DIFF
--- a/tests/pretty.rs
+++ b/tests/pretty.rs
@@ -261,11 +261,15 @@ fn table_array() {
     assert_eq!(toml, &result);
 }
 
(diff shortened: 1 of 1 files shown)


vitiral:
it would be great if a new version could be pushed after this (can be a micro bump)

alexcrichton:
Will do!

PR #220, opened 2017-12-02
Small copy-paste error in license text

LeopoldArkham:
Copied my license text from you and realized it was referencing Serde. Don't mess with Big License!


alexcrichton:
Thanks!

alexcrichton:
> Don't mess with Big License!

I've learned at this point to never mess with 'Big $anything'

PR #221, opened 2018-01-07
change the Error & ErrorInner struct to public

dvarum12:
this will make it possible to use them for error messages and debugging

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -76,21 +76,21 @@ pub fn from_str<'de, T>(s: &'de str) -> Result<T, Error>
 /// Errors that can occur when deserializing a type.
 #[derive(Debug, Clone)]
 pub struct Error {
-    inner: Box<ErrorInner>,
+    pub inner: Box<ErrorInner>,
 }
 
 #[derive(Debug, Clone)]
-struct ErrorInner {
-    kind: ErrorKind,
-    line: Option<usize>,
-    col: usize,
-    message: String,
-    key: Vec<String>,
+pub struct ErrorInner {
+    pub kind: ErrorKind,
+    pub line: Option<usize>,
+    pub col: usize,
+    pub message: String,
+    pub key: Vec<String>,
 }
 
 /// Errors that can occur when deserializing a type.
 #[derive(Debug, Clone)]
-enum ErrorKind {
+pub enum ErrorKind {
     /// EOF was reached when looking for a value
     UnexpectedEof,
 


alexcrichton:
Thanks for the PR! Can you elaborate on what you wanted to do with these? Ideally this'd be exposed through methods rather than public fields which makes it difficult to change the struct in the future, but I'd just want to make sure we exposed the methods necessary!

dvarum12:
Mainly I want to use it for creating better error messages for then something wrong in the config file. Right now as you can see I only handles a error, but I am just adding to the code as I encounter new errors. This is why I want the structs Error & ErrorInner to be public. If I complete the error handler, my plan is to commit it, but right now I don't know then or if it will happen.
Hope this help understanding the thoughts behind my PR 😄 

```
fn error_handler<T>(config: serde::export::Result<T, toml::de::Error>) -> T where T: self::serde::export::fmt::Debug {
    if config.is_err() {
        let error = &config.unwrap_err().inner;
        match error.kind {
            toml::de::ErrorKind::Custom => {
                println!("You have to add the {}, in the section [{}] of the config file", error.message, error.key.first().unwrap());
            }
            _ => {
                println!("-=( Un-handled error )=-");
                println!("{:?}", error);
            }
        }
        ::std::process::exit(1);
    }
    config.unwrap()
}
```

alexcrichton:
Ok thanks! Would it be possible to add accessors through methods instead of making the internals public though? Perhaps just for the information you're interested in for this?

dvarum12:
I guess you could create a "get" method for all the values, but I assume that I am gonna use all the value at some point (but don't know) so don't know if it make sense to do that?
I can also try to implement a method there returns my error message and then there is a error with a error message, it will just panic and then we can slowly implement error messages for all the cases, but again that is kind of a compromise.

alexcrichton:
Hm ok. Do you understand why though structures are in general private by default? I think it'd be ok to expose more information, but I wouldn't want to expose *too* much because it may be difficult to maintain the information over time.

dvarum12:
Yes I understand, but I still access to this information (somehow) and I assume other people do to, this was why I suggested to make them public. However if you have a better solution or a plan with the private structures then I am all ears, but leaving them as they are without using them are not the solution, if you understand what I mean.

dvarum12:
Yes, you could create some method to expose the values that I need, but I still think that it is better to expose the error structures, considering that there is not real error handling in the toml-rs, but again that is up to you. I just think it is better to make the errors public until there have been implement some sort of error handling into toml-rs, because this way people them self can implement there own error handling and maybe someone will create a really good one that can be implement into toml-rs.

alexcrichton:
Yes I'd personally prefer to make methods and accessors rather than making structures/enums public.

dvarum12:
Okay, but can I get you to create methods that exposes these error values?

alexcrichton:
Er sorry for being late to respond, but to clarify for sure yeah I'm fine adding accessors to learn about the structure of each error. Otherwise though I don't think we'll be taking this particular strategy, so I'm going to close this.

PR #223, opened 2018-01-22
add extraction by path

Fiedzia:
This was very useful for me when dealing with toml configuration files

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -90,6 +90,54 @@ impl Value {
         index.index_mut(self)
     }
 
+    /// Walk through nested values according to path
+    /// for example if the path is "foo.bar", it will extract foo key, and then bar key,
+    /// if it is there
+    pub fn get_path(&self, path: &str) -> Option<&Value> {
+        let mut value: &Value = &self;
+        for path_part in path.split('.') {
+            value = match value {
+                &Value::Table(ref t) => match t.get(path_part) {
+                    Some(v) => &v,
+                    None => return None,
+                },
+                _ => return None,
+            };
+        }
+        Some(value)
+    }
+
+    /// Extracts array value by providing path
+    pub fn get_array_from_path(&self, path: &str) -> Option<&Vec<Value>> {
+        self.get_path(path).and_then(|v| v.as_array())
+    }
+
+    /// Extracts boolean value by providing path
+    pub fn get_bool_from_path(&self, path: &str) -> Option<bool> {
+        self.get_path(path).and_then(|v| v.as_bool())
+    }
+
+    /// Extracts str value by providing path
+    pub fn get_str_from_path(&self, path: &str) -> Option<&str> {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! This crate actually used to have this method long ago but it ended up getting removed in the transition to serde. One problem historically though was that splitting on `.` doesn't account for TOML parsing of keys like `foo.'bar.baz'`, and handling cases like that is often quite difficult. I wonder though if there's perhaps another way to tackle this problem?

Fiedzia:
> One problem historically though was that splitting on . doesn't account for TOML parsing of keys like foo.'bar.baz'

I could make parser more sophisticated if needed, but in my case I was dealing with files under my control (app configuration). Is there any spec or convention for rules of key parsing?

If there isn't, I'd say just leave it as is, if someone has special needs, they will require custom parsing anyway.

alexcrichton:
AFAIK there's not a great spec for something like this, but how about this:

* Let's not include this because it can have be a footgun
* However, perhaps we can include this as an example/documentation in the repo?

That'd at least make it easy to copy/paste into a local project!

Fiedzia:
I'd rather either figure out what the parsing rules should be, allow to use custom parser or move it to some toml-utils crate if you think it doesn't belong here.

alexcrichton:
There was a historical implementation of this in previous versions of the crate which may help?

Fiedzia:
If I understand it right, quoted text in case foo.'bar.baz' should follow the rules of basic strings, for example path to
[foo]
"a\"b" = 3
would be written as foo."a\"b", and all I'd need to do would be to call Tokenizer::basic_string on anything in quotes, and that would cover all cases. This would be very simple to do, match the spec (to the extent basic_string does). Nowhere near complexity of old parsing code (https://github.com/alexcrichton/toml-rs/commit/f66d8bcf33530c858a502bfa170f2383a8cbc204#diff-2c09afcdc3c420ab0678ba9b5e83959c)

alexcrichton:
I think perhaps yeah? I'd still personally prefer this to not be in the main repo though in the sense that "access paths" aren't necessarily a standard feature related to TOML

remLse:
I assume the reason something like this is not in the spec is because keys containing dots are allowed, in which case this code does not behave as expected. [Example from the TOML spec](https://toml.io/en/v1.0.0#keys):
```
name = "Orange"
physical.color = "orange"
physical.shape = "round"
site."google.com" = true
```
JSON:
```
{
  "name": "Orange",
  "physical": {
    "color": "orange",
    "shape": "round"
  },
  "site": {
    "google.com": true
  }
}
```
Simply splitting on all dots would treat "google" as a table with a key "com", which is not the case.

That means this code works fine for people who know they're only going to deal with keys without dots, but it won't work reliably for any valid TOML file.

epage:

PR #229, opened 2018-03-25
[WIP] Nested structs fail to serialize with ValueAfterTable error because struct properties are not ordered

klausi:
Problem: if you use nested structs and you specify primitive struct properties after the nested struct then serialization fails.

Attached is a test case that demonstrates the problem.

Not sure how and where to solve this exactly, maybe you can give me some pointers where to start? I think we need to order struct properties before handing them down to the next serialization step, so that primitive properties always come first.

cc @dtolnay who might also have an idea about this :-)


epage:

Issue #231, opened 2018-04-05
inserting values with [ ]


jaroslaw-weber:
I was able to insert new key/value pairs to json with [] indexing, but it seems is not possible with toml crate. Would the implementation is difficult? How would I go about adding new information into Value::Table?

```
extern crate toml;
use toml::value::Value as Toml;
use toml::Value;

fn main() {

    let mut toml_obj = "foo = 'bar'".parse::<Value>().unwrap();
    toml_obj["new_key"]=Toml::Boolean(true);
    println!("{:?}", &toml_obj);
}
```

alexcrichton:
Thanks for the report! I believe this works in Serde because `IndexMut` returns a reference to `null` if the key isn't otherwise present. Unfortunately TOML doesn't have the same notion of null so the same trick doesn't work for us.

In that sense this won't be easily implementable until we get something like `IndexSet` as a trait in the language :(

jaroslaw-weber:
@alexcrichton  
hey, thanks for response! What would be a way to insert a new value into Value?
It would be great to have 
```
Value::insert(&mut self, key: &str, value: Value)-> Result<(), Error>
```
for modifying the Value. Would it be possible to implement?

alexcrichton:
Certainly! That sounds plausible to me

ehuss:
cc #193 which is similar.

epage:
Maintenance of this crate has moved to the https://github.com/toml-rs/toml repo.  As a heads up, we plan to move `toml` to be on top of `toml_edit`, see https://github.com/toml-rs/toml/issues/340.

Closing this out.  If this is still a problem, feel free to recreate this issue in the new repo.

PR #233, opened 2018-04-12
adding function to insert new values to toml table

jaroslaw-weber:
I implement inserting new values directly into the Value.
I wrote about it in an Issue here:
https://github.com/alexcrichton/toml-rs/issues/231

We can already quickly get values from Value with index access, but inserting the new Value is kind of clunky, so I added the helper function directly inside the Value.

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -204,6 +204,14 @@ impl Value {
             Value::Table(..) => "table",
         }
     }
+
+    /// Adding new Value to a Table. Returns `None` when fail (Value is not a table);
+    pub fn insert(&mut self, key: &str, value: Value) -> Option<()>
+    { 
+        let table = self.as_table_mut()?;
+        table.insert(key.to_string(), value);
+        Some(())
+    }
 }
 
 impl<I> ops::Index<I> for Value where I: Index {
--- a/test-suite/tests/invalid-misc.rs
+++ b/test-suite/tests/invalid-misc.rs
@@ -1,4 +1,5 @@
 extern crate toml;
+use toml::Value;
 
 #[test]
 fn bad() {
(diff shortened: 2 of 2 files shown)


dtolnay (changes requested):
I would prefer not to have this method and instead recommend that people manipulate `toml::value::Table` rather than `Value` if they want to write code like this. I believe the code in your test case is clearer if written using Table and not Value.

```rust
let mut t: Table = toml::from_str("a=1").expect("failed to create Table");
t.insert("b".to_owned(), Value::Integer(2));
t.insert("c".to_owned(), Value::Integer(3));
```

jaroslaw-weber:
@dtolnay 
hey, thanks for the feedback! I understand your point of view, writing `as_mut_table().unwrap().insert(...)` is not that much longer and there is a way to insert a value already so it is not that necessary. Still, I miss inserting values with index just like in serde. I think it is common operation and it is worth to consider adding inserting directly in the Value.

dtolnay:
Maybe I have not used the toml crate enough myself to have hit this use case -- could you point out an example of real code that would benefit from having Value::insert?

jaroslaw-weber:
@dtolnay 
I edited my post, please read it again!
I was using serde json before and was confused why cannot I directly inserting things into Value (in json you can insert to value with index) and was doing it like this:
```
let mut t: Value = toml::from_str("a=1").expect("failed to create Value");
t.as_mut_table().unwrap().insert("b".to_owned(), Toml::Integer(1));
```
and this was little long so I thought we could use inserting new Value directly in the Value.

But toml crate is little different, I didn't know you are supposed to do:
```
let mut t: Table = toml::from_str("a=1").expect("failed to create Table");
```

So I agree with you, we don't need adding insert method in the Value. But I would add inserting example to readme/docs, because it is quite common pattern and people who used serde json may find it little confusing.

alexcrichton:
Adding some examples to the documentation sounds great to me as well!

epage:

Issue #236, opened 2018-04-29
Support emitting spans when deserializing structures


udoprog:
Hey,

I have a use-case where I perform higher-level diagnostics on top of existing structures, and in order to provide nice errors to the user I need to have access to the span information where a given structure was deserialized _from_.

Following is a somewhat naive proposal, since I'm not too familiar with the low-level details of how to implement a deserializer for serde.

One option could be to include spans in the tokens returned by the tokenizer, then introduce a generic wrapper (e.g. `Spanned<T>`) that includes the spans when de-serializing so that they can be used in later errors.

```rust
struct Span {
  /// Byte offset where the span starts.
  pub start: usize,
  /// Byte offset where the span ends.
  pub end: usize,
}

struct Spanned<T> {
  pub span: Span,
  pub value: T,
}

#[derive(Deserialize)]
pub struct Foo {
  bar: Spanned<String>,
}
```

For this to work with `toml::Value`, the spans would have to be included here, which might exclude it for backwards compatibility. In that case a new set of structures (e.g. `toml::SpannedValue`) could be introduced.

alexcrichton:
Would be nice to do this for sure! This crate doesn't do a great job in general about capturing source information for deserialized items, but I'd be more than willing to accept PRs to add support!

udoprog:
I am fairly certain that the serde-based solution will have a hard dependency on something like serde-rs/serde#1184, the existing deserialization facilities provided by serde currently does not permit communicating spans. @dtolnay?

dtolnay:
The existing deserialization facilities provided by serde currently "does not permit" communicating datetimes or arbitrary precision numbers, but this crate can deserialize datetimes and serde_json can deserialize arbitrary precision numbers. You just need to figure out how you would want span information represented in the existing Serde data model.

I believe serde-rs/serde#1184 is unrelated because it only involves the error case.

If possible, the easier solution to https://github.com/reproto/reproto/issues/48 would be to perform the high-level diagnostics during deserialization rather than after deserialization has succeeded.

dtolnay:
Here is a quick proof of concept of what I mean by representing span information in the existing Serde data model: https://github.com/alexcrichton/toml-rs/pull/238/files.

```rust
#[macro_use]
extern crate serde_derive;

extern crate toml;
use toml::spanned::Spanned;

#[derive(Deserialize)]
struct Udoprog {
    s: Spanned<String>,
}

fn main() {
[...]

PR #237, opened 2018-05-01
Make Tokenizer emit spans

udoprog:
This is the first step at addressing #236.

This modifies the tokenizer to emit span information for each token as well, and does the necessary modifications to deserialization to keep things working for now.


alexcrichton:
Looks great!

PR #239, opened 2018-05-07
Support spans when deserializing serde structures

udoprog:
This PR introduces Spanned based on the previous POC by @dtolnay.

Internally we deserialize spans as "intermediate structs" with special naming conventions, which communicates to the `Deserializer` that it should emit a Map with similarly special names that we can pick up.

This PR contains a couple of changes:

* Introduces a `Spanned<T>` type that can be used in structs to encapsulate spans.
* Unified the naming convention of constants for `datetime` and `spanned`.
* Added `eat_spanned` and `expect_spanned` to `Tokenizer` to correctly parse tables and arrays. 

I'm unsure it is complete, the infrastructure needed to support Datetime permeates more of the deserialize infrastructure, but I haven't been able to decipher _why_ yet. Any feedback is appreciated.

#### Future work

Ironically this doesn't help me with my original problem, but it's a fair step down the road. I use `toml::Value` in a similar way that `Cargo` uses it to permit more flexible decoding than is afforded through databinding. Introducing spans into the existing `toml::Value` would not be a backwards compatible change, so the only alternative I see is to have separate types (e.g. `toml::SpannedValue`) which contains the necessary spans. Suggestions are welcome.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ travis-ci = { repository = "alexcrichton/toml-rs" }
 
 [dependencies]
 serde = "1.0"
+serde_derive = "1.0"
 
 [dev-dependencies]
-serde_derive = "1.0"
 serde_json = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,4 +22,5 @@ travis-ci = { repository = "alexcrichton/toml-rs" }
 serde = "1.0"
 
 [dev-dependencies]
+serde_derive = "1.0"
 serde_json = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,6 @@ travis-ci = { repository = "alexcrichton/toml-rs" }
 
 [dependencies]
 serde = "1.0"
-serde_derive = "1.0"
 
 [dev-dependencies]
 serde_json = "1.0"
--- a/src/de.rs
+++ b/src/de.rs
@@ -821,16 +821,20 @@ impl<'a> Deserializer<'a> {
                 Value { e: E::Boolean(false), start: start, end: end }
             }
             Some((span, Token::Keylike(key))) => self.number_or_date(span, key)?,
-            Some((_, Token::Plus)) => self.number_leading_plus()?,
-            Some((Span { start, end }, Token::LeftBrace)) => {
-                self.inline_table().map(|table| Value {
+            Some((span, Token::Plus)) => self.number_leading_plus(span)?,
+            Some((Span { start, .. }, Token::LeftBrace)) => {
+                self.inline_table().map(|(Span { end, .. }, table)| Value {
                     e: E::InlineTable(table),
                     start: start,
                     end: end
                 })?
             }
-            Some((Span { start, end }, Token::LeftBracket)) => {
-                self.array().map(|array| Value { e: E::Array(array), start: start, end: end })?
+            Some((Span { start, .. }, Token::LeftBracket)) => {
+                self.array().map(|(Span { end, .. }, array)| Value {
+                    e: E::Array(array),
+                    start: start,
+                    end: end
+                })?
             }
             Some(token) => {
                 return Err(self.error(at, ErrorKind::Wanted {
@@ -882,11 +886,13 @@ impl<'a> Deserializer<'a> {
         }
     }
 
-    fn number_leading_plus(&mut self) -> Result<Value<'a>, Error> {
-        let start = self.tokens.current();
+    fn number_leading_plus(&mut self, Span { start, .. }: Span) -> Result<Value<'a>, Error> {
+        let start_token = self.tokens.current();
         match self.next()? {
-            Some((span, Token::Keylike(s))) => self.number(span, s),
-            _ => Err(self.error(start, ErrorKind::NumberInvalid)),
+            Some((Span { end, .. }, Token::Keylike(s))) => {
+                self.number(Span { start: start, end: end }, s)
+            },
+            _ => Err(self.error(start_token, ErrorKind::NumberInvalid)),
         }
     }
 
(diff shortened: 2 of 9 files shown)


udoprog:
Note: I'm also pushing it through the build system to check for 1.15 compatibility. Initial builds are probably gonna break.

dtolnay:
Would it be possible to support spans in a map key? `BTreeMap<Spanned<String>, Spanned<String>>`

Cargo.toml
```diff
@@ -20,7 +20,7 @@ travis-ci = { repository = "alexcrichton/toml-rs" }
 
 [dependencies]
 serde = "1.0"
+serde_derive = "1.0"
```
dtolnay: Looks like you wrote out the Deserialize impl so this is no longer used.

udoprog:
@dtolnay I was fiddling with some tests for cases which currently does not work. Pushed them out in dffb9524acb25aa679687c9bfcbd6b164b5437da.

src/spanned.rs
```diff
@@ -0,0 +1,153 @@
...
+            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+                formatter.write_str("a TOML spanned")
+            }
+
+            fn visit_map<V>(self, mut visitor: V) -> Result<Spanned<T>, V::Error>
+                where V: de::MapAccess<'de>
+            {
+                let start = visitor.next_key::<StartKey>()?;
+
+                if start.is_none() {
```
dtolnay: I would recommend cutting out the `key_deserialize` macro and using:

```rust
if visitor.next_key()? != Some(START) {
```

src/de.rs
```diff
@@ -594,6 +606,46 @@ impl<'de> de::IntoDeserializer<'de, Error> for Value<'de> {
...
+
+impl<'de> de::MapAccess<'de> for SpannedDeserializer<'de> {
+    type Error = Error;
+
+    fn next_key_seed<K>(&mut self, seed: K) -> Result<Option<K::Value>, Error>
+    where
+        K: de::DeserializeSeed<'de>,
+    {
+        if self.start.is_some() {
+            seed.deserialize(spanned::START.into_deserializer()).map(Some)
```
dtolnay: This needs to be:

```rust
use serde::de::value::BorrowedStrDeserializer;

seed.deserialize(BorrowedStrDeserializer::new(spanned::START)).map(Some)
```

in order for my other comment to work.

udoprog:
Regarding `BTreeMap<Spanned<String>, Spanned<String>>`, what's the identity of `Spanned<T>`? If it's transitive to `T`, two differently spanned T's would override each other as expected. On the other hand, two differently spanned `T`s aren't equal.

Generally I address this by distinguishing between the key of a structure and its span, like: `BTreeMap<String, (Span, T)>`. But that isn't really an option here.

dtolnay:
I would expect `Map<Spanned<K>, V>` to behave exactly like `Map<K, V>` just with more information captured during deserialization.

src/spanned.rs
udoprog: Fixed, thanks!

src/de.rs
udoprog: Fixed, thanks!

Cargo.toml
udoprog: Fixed, thanks!

test-suite/tests/spanned.rs
```diff
@@ -0,0 +1,43 @@
...
+    good::<HashMap<String, u32>>(
+        "foo = {\"foo\" = 42, \"bar\" = 42}",
+        "{\"foo\" = 42, \"bar\" = 42}"
+    );
+    // array
+    good::<Vec<u32>>(
+        "foo = [0, 1, 2, 3, 4]",
+        "[0, 1, 2, 3, 4]"
+    );
+    // datetime
```
dtolnay: I tried an actual `Datetime` but it seems to get the wrong span. Is this a tokenizer issue?

```rust
good::<Datetime>("foo = 1997-09-09T09:09:09Z", "1997-09-09T09:09:09Z");
```

```
---- test_spanned_field stdout ----
	thread 'test_spanned_field' panicked at 'assertion failed: `(left == right)`
  left: `26`,
 right: `19`', tests/spanned.rs:20:9
```

udoprog: Fixed in 2b40d54, thanks!

udoprog:
> I would expect Map<Spanned<K>, V> to behave exactly like Map<K, V> just with more information captured during deserialization.

I half-agree with this, but it becomes awkward when trying to implement and _use_: https://play.rust-lang.org/?gist=317592575bffab2bb63bc89ef9a3f9a0&version=stable&mode=debug (possibly my impl-fuu is not strong enough, so please take a stab at this).

But if I have a map which includes spans, I'd want to be able to:
* Have a way to get/remove elements by a reference of the contained value in the key (`&K` in `Map<Spanned<K>, V>`).
* Have a way to access the span of a given key, by a reference of the contained value in the key (get `Span` from `&K` in `Map<Spanned<K>, V>`).

Note: One approach that satisfies this is to use `Map<K, (Span, V)>`, where the first element in the tuple is the _span_. We could introduce a third type specialized for map-serialization to support this: `Map<K, MapSpanned<V>>`.

src/lib.rs
```diff
@@ -169,3 +169,7 @@ mod tokens;
 
 #[doc(hidden)]
 pub mod macros;
+
+pub mod spanned;
```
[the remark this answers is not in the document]
udoprog: Will do. Thanks.

src/spanned.rs
```diff
@@ -0,0 +1,106 @@
...
+
+/// A spanned value, indicating the range at which it is defined in the source.
+#[derive(Debug)]
+pub struct Spanned<T> {
+    /// The start range.
+    pub start: usize,
+    /// The end range (exclusive).
+    pub end: usize,
+    /// The spanned value.
+    pub value: T,
```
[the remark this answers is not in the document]
udoprog: These are the set of methods I'm fiddling with right now:

```rust
impl<T> Spanned<T> {
    fn span(&self) -> (usize, usize) { .. }

    fn borrow(&self) -> &T { .. }

    fn take(self) -> T { .. }
}
```

Got an opinion on this?

src/spanned.rs
```diff
@@ -0,0 +1,137 @@
...
+        self.end
+    }
+
+    /// Access the span for the spanned value.
+    pub fn span(&self) -> (usize, usize) {
+        (self.start, self.end)
+    }
+
+    /// Take the inner value.
+    pub fn take(self) -> T {
```
alexcrichton: For consistency with libstd, mind naming this `into_inner`? I think we may want to go ahead and add methods like `get_ref` and `get_mut` here as well

udoprog: Done, thanks!

PR #241, opened 2018-05-08
Treat unicode hex digits case-insensitively

### LINKED ISSUE #240: Unicode escapes only accept upper case hex digits
For the input:
```
foo = " \uf04b "
```
I was receiving a puzzling error:
```
invalid hex escape character in string: `f`
```
TOML is specified as ABNF:
```
escape-seq-char =/ %x75 4HEXDIG ; uXXXX                U+XXXX
escape-seq-char =/ %x55 8HEXDIG ; UXXXXXXXX            U+XXXXXXXX

HEXDIG = DIGIT / "A" / "B" / "C" / "D" / "E" / "F"
```

According to RFC 5234:
>     ABNF strings are case insensitive and the character set for these
>      strings is US-ASCII.
>
>   Hence:
>
>         rulename = "abc"
>
>   and:
>
>         rulename = "aBc"
>
>   will match "abc", "Abc", "aBc", "abC", "ABc", "aBC", "AbC", and
>   "ABC".

So, `\uf04b` and `\uF04B` really should mean the same thing.

avdv:
Fixes #240.

This is my first try at writing some Rust, so bear with me. I am still struggling with writing a test case...

### DIFF
--- a/src/tokens.rs
+++ b/src/tokens.rs
@@ -2,6 +2,7 @@ use std::borrow::Cow;
 use std::char;
 use std::str;
 use std::string;
+use std::string::String as StdString;
 
 use self::Token::*;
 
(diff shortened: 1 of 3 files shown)


alexcrichton:
Awesome, thanks! You can probably copy/paste some existing test cases [such as these](https://github.com/alexcrichton/toml-rs/blob/master/test-suite/tests/valid.rs#L219-L221) to ensure that they all match up.

alexcrichton:
Also it looks like `is_ascii_hexdigit` was added pretty recently, mind inlining the definition here?

avdv:
> Also it looks like is_ascii_hexdigit was added pretty recently, mind inlining the definition here?

How about using `char::is_digit` instead, that seems to be available in Rust since version 1.0.0?!

alexcrichton:
Sure yeah! You'll want to validate as well that it's ascii (as that `is_digit` is "is a unicode digit") but that shouldn't be too hard

avdv:
I have updated the PR, while adding some test cases.

alexcrichton:
Awesome, thanks!

Issue #224, opened 2018-02-05
Checklist for new TOML updates


ehuss:
Here is a list of changes to the TOML spec that have not yet been added to toml-rs.

I'm wondering if you would prefer to wait until the spec is finalized to a specific version, or if you would prefer to be proactive in keeping up-to-date.

- [x] Support hex, octal, and binary integer formats ([commit](https://github.com/toml-lang/toml/commit/c69875f7b7d3a764ac8699fbd7915d963a25685c)).
- [x] Support special float values inf and nan, including signs [#506](https://github.com/toml-lang/toml/pull/506).
- [x] Support space instead of T for datetime objects [#501](https://github.com/toml-lang/toml/pull/501).
- [x] Bare U+007f should be rejected from strings [#485](https://github.com/toml-lang/toml/issues/485).  
- [x] Support dotted keys [#505](https://github.com/toml-lang/toml/pull/505).

ehuss:
@alexcrichton Tom [released 0.5 today](https://github.com/toml-lang/toml/releases/tag/v0.5.0), I was wondering if you would be interested in PRs for the new features and changes?

alexcrichton:
I would definitely be interested! If they come with breaking API changes to the `toml` crate I'd be totally down for a major version bump as well

ehuss:
@alexcrichton Hm, I completely forgot about updating the macros until after I pushed.  I will look at supporting them tomorrow, but I often struggle with macros, so it may take a little time.

alexcrichton:
Ah yeah no worries, it's fine to update them as-needed.

birkenfeld:
Should the README be updated to say that 0.5 is (at least partially) now supported?

alexcrichton:
Er oops, yes indeed!

PR #252, opened 2018-07-11
0.5: Support dotted keys.

ehuss:
cc #224

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -1179,12 +1179,12 @@ impl<'a> Deserializer<'a> {
             values.push((key, value));
             return Ok(());
         }
-        match values.iter_mut().find(|(k, _)| *k == key) {
-            Some((_, Value { e: E::InlineTable(ref mut v), .. })) => {
+        match values.iter_mut().find(|&&mut (ref k, _)| *k == key) {
+            Some(&mut (_, Value { e: E::InlineTable(ref mut v), .. })) => {
                 return self.add_dotted_key(key_parts, value, v);
             }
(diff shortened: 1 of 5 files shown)


alexcrichton:
Thanks! I wouldn't worry too much about errors and spans, we can always try to fix those up later I think. Should it be possible to use these keys to insert into normal tables as well though? Something like:

```toml
a.b = 1
[a.c]
d = 3

ehuss:
Good question.  There's a lot of discussion on the toml issues about when tables should be open/closed, but it doesn't look like there is much consensus. I'll ask them about that case.

What do you think about keeping it conservative for now, and expanding support later if that's the direction TOML wants to go?

I'd also be fine with delaying this PR until they decide on these issues.  This is by far the biggest change, and probably going to be controversial to some.

alexcrichton:
I'd be fine with either the route of holding off until we hear back on the issue or adding the conservative pieces here. We could also add support but require a flag to turn it on in the parser perhaps?

ehuss:
I'm keen to wait for clarity on the spec.  I'll circle back once I have a better understanding how it should work.

ehuss:
I've pushed a small change to add a further restriction that dotted keys can't insert into previous inline tables.

I've been thinking about this for a while.  I feel like it would be best to initially be conservative and don't allow the three table types to intermix.  The limitations can be removed later if there are compelling reasons or use cases.  I think it will be easier to relax the restrictions later than to initially start with something too permissive.

alexcrichton:
Sounds great to me, thanks!

PR #254, opened 2018-08-05
Support the new 0.5 syntax in the toml! macro.

ehuss:
- Space in datetime.
- +- inf, nan, 0.0
- Dotted keys.


ehuss:
It's not exactly pretty, but I'm not a macro wizard and it works.  Maybe in the future something like proc macro can be used instead?

PR #255, opened 2018-08-05
Support deserializing from other integer types

kornelski:
Fixes https://github.com/serde-rs/json/issues/471

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -423,6 +423,22 @@ impl<'de> de::Deserialize<'de> for Value {
                 Ok(Value::Integer(value))
             }
 
+            fn visit_u64<E: de::Error>(self, value: u64) -> Result<Value, E> {
+                if value <= i64::max_value() as u64 {
+                    Ok(Value::Integer(value as i64))
+                } else {
+                    Err(de::Error::custom("u64 value was too large"))
+                }
+            }
+
+            fn visit_u32<E>(self, value: u32) -> Result<Value, E> {
+                Ok(Value::Integer(value.into()))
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks! Could a test be added for this as well?

kornelski:
I've ran into this via `serde_json`, but I'm not sure if you want to have it as a dev dependency.  I'm not familiar enough with serde's internals to mock it out.

alexcrichton:
Ah sure yeah, adding such a dep seems fine by me!

kornelski:
Ok, added test.

Michael-F-Bryan:
I just ran into this with `mdbook` and `mdbook-linkcheck` (in particular, [`mdbook::config::Search`]). 

It looks like the last release for `toml` was in April, is there any chance a patch release could be added so we have access to this?

[`mdbook::config::Search`]: https://docs.rs/mdbook/0.2.1/mdbook/config/struct.Search.html

alexcrichton:
Ah sorry about that, published now!

ehuss:
@alexcrichton The 0.4.7 release brings in a lot of changes compared to 0.4.6, including the new 0.5 syntax. I'm concerned that may bring in unexpected changes to people depending on "0.4". Is that OK?

alexcrichton:
@ehuss ah true, I forgot about that! I do think though that we'll probably want to stick to the 0.4 chain if we can (as it's all "largely backwards compatible" anyway I believe)

If any problems are encountered though we can republish as 0.5

PR #257, opened 2018-08-12
Add from_reader and to_writer convenience methods

### LINKED ISSUE #215: Add `from_reader` method
Like serde_json, it will be nice if we have `from_reader`.

If we don't have one...
```rust
use std::fs::File;
use std::io::Read;

let mut file = File.open(path)?;
let mut content = String::new();
file.read_to_string(&mut content)?;
let some_struct: SomeStruct = toml::from_str(&content)?;
```

But if we have one...
```rust
use std::fs::File;

let file = File.open(path)?;
let some_struct: SomeStruct = toml::from_reader(&file)?;
```

I'm just beginner of rust, so I don't have enough ability to make it myself and make a pull request. But if anyone can make, codes will be cleaner like above.

### LINKED ISSUE #216: Why can't you encode to any "Write"?
If i grep for `self.dst` in `ser.rs`, all I can find are `push_str` and `write!` calls, so it looks like any writer should be accepted. Would a pull request be considered?

Edit: Also, this change would mean that a string wouldn't need to be created on every `<toml::Value as fmt::Display>::fmt` call.

RazerM:
Resolves #215, resolves #216.

I used `Arc<io::Error>` as suggested in #215.

### DIFF
--- a/src/ser.rs
+++ b/src/ser.rs
@@ -29,6 +29,7 @@
 use std::cell::Cell;
 use std::error;
 use std::fmt::{self, Write};
+use std::io;
 use std::marker;
 use std::rc::Rc;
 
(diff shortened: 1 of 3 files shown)


src/ser.rs
```diff
@@ -94,6 +95,60 @@ pub fn to_string<T: ?Sized>(value: &T) -> Result<String, Error>
...
+///     toml::to_writer(&mut file, &config).unwrap();
+/// }
+/// ```
+pub fn to_writer<W, T: ?Sized>(writer: &mut W, value: &T) -> Result<(), Error>
+where
+    W: io::Write,
+    T: ser::Serialize,
+{
+    let b = to_vec(value)?;
+    writer.write(&b).unwrap();
```
dtolnay: This needs to use [`write_all`](https://doc.rust-lang.org/std/io/trait.Write.html#method.write_all) or else it may silently drop part of the data.

dtolnay: Also instead of unwrap it should construct the right toml Error.

RazerM: Sorry I did this a while ago and forgot I didn't finish it. What's the best way to fit `io::Error` into `toml::ser::Error` which implements `Eq`?

src/ser.rs
```diff
@@ -94,6 +95,60 @@ pub fn to_string<T: ?Sized>(value: &T) -> Result<String, Error>
...
+///     toml::to_writer(&mut file, &config).unwrap();
+/// }
+/// ```
+pub fn to_writer<W, T: ?Sized>(writer: &mut W, value: &T) -> Result<(), Error>
+where
+    W: io::Write,
+    T: ser::Serialize,
+{
+    let b = to_vec(value)?;
+    writer.write_all(&b).unwrap();
```
alexcrichton: Could this use `?` instead of `unwrap()`?

alexcrichton:
Looks great to me, just one minor nit. Thanks!

src/ser.rs
RazerM: I commented in the "outdated" review above:

> I did this a while ago and forgot I didn't finish it. What's the best way to fit `io::Error` into `toml::ser::Error` which implements `Eq`?

alexcrichton: Ah sorry missed that, perhaps instead of that this function could return `io::Error`?

Arignir:
Bump! Would be great to have this in `toml-rs` :)

bbqsrc:
2020 bump life

RazerM:
If anyone wants to continue where I left it off, `to_writer` returns `toml::ser::Error` which can't contain `io::Error` and I didn't know how to fix that.

pickfire:
I believe as for that, we could add `IoError(std::io::Error)` (or something else) to https://github.com/alexcrichton/toml-rs/blob/7d8ed33c949dfff9df873193ac18183fc136925f/src/ser.rs#L164-L206

Then, I believe can just use `writer.write_all(&b).map_err(Error::IoError)?`, not tested but I think it is the rough idea.

bbqsrc:
It would make sense to just replicate whatever `serde_json` does in the same situation.

pickfire:
serde_json uses trait to handle deserialization. `to_writer` is used as a base for other methods.

jmfiaschi:
Tell me if this fix it's OK for you : https://github.com/RazerM/toml-rs/pull/1

ijackson:
A possibility would be to return `std::io::Error`.  That can already contain other errors as an `inner error`.

WhyNotHugo:
I've rebased this onto current `master` in #452.

> If anyone wants to continue where I left it off, to_writer returns toml::ser::Error which can't contain io::Error and I didn't know how to fix that.

I removed the `PartialEq` and `Eq` traits from `ErrorKind`, and that unblocked that issue. Tests pass, and some very basic testing _seems_ to work, but I'm unaware if that'll have any larger consequences.

epage:

PR #261, opened 2018-09-25
Fix multi-line strings are not allowed for key (#260)

### LINKED ISSUE #260: Key should not allow multi-line strings
## Description

In spec https://github.com/toml-lang/toml#keys

Quoted keys are clarified as

> the exact same rules as either basic strings or literal strings

TOML clearly distinguishes basic string and multi-line basic string (literal string is also).

https://github.com/toml-lang/toml#string

So table key and quoted key should not allow multi-line basic string and multi-line literal string.

ABNF definition also describes that.

https://github.com/toml-lang/toml/blob/master/toml.abnf

```
string = ml-basic-string / basic-string / ml-literal-string / literal-string

quoted-key = basic-string / literal-string
```

`string` contains `ml-*` but `quoted-key` doesn't.

## Repro

```rust
extern crate toml;

use toml::Value;

fn main() {
    let source = "'''foo''' = 42";
    let v: Value = source.parse().unwrap();
    println!("{:?}", v);

    let source = "['''foo''']";
    let v: Value = source.parse().unwrap();
    println!("{:?}", v);
}
```

I expected parser to raise an error but got an input as follows:

```
Table({"foo": Integer(42)})
Table({"foo": Table({})})
```

rhysd:
Fixes #260

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -152,6 +152,9 @@ enum ErrorKind {
     /// An empty table key was found.
     EmptyTableKey,
 
+    /// Multiline strings are not allowed for key
+    MultilineStringKey,
+
     /// A custom error which could be generated when deserializing a particular
     /// type.
     Custom,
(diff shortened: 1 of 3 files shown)


alexcrichton:
Awesome, thanks!

Issue #225, opened 2018-02-09
externally tagged enums


dherman:
When you define an enum and derive `Serialize` or `Deserialize`, you might expect it to get the ["externally tagged"](https://serde.rs/enum-representations.html) semantics described by the serde documentation -- where each variant is represented as a table with a key named after the variant name. But AFAICT this representation doesn't work. Here's a simple example, a config format with a `[main]` section and a `plugin` field that can be one of `{ "url" = ...}` or `{ "bin" = ... }`:

```rust
extern crate toml;
#[macro_use]
extern crate serde_derive;
extern crate serde;

#[derive(Serialize, Deserialize, Debug)]
pub struct Config {
    pub main: Option<Main>
}

#[derive(Serialize, Deserialize, Debug)]
#[serde(rename = "main")]
pub struct Main {
    pub plugin: Option<Plugin>,
}

#[derive(Serialize, Deserialize, Debug)]
#[serde(rename_all = "lowercase")]
pub enum Plugin {
    Url(String),
    Bin(String)
}

const SRC: &'static str = "
[main]
plugin = { \"bin\" = \"echo 'hello'\" }
";

fn main() {
    match toml::from_str::<Config>(SRC) {
        Err(e) => { eprintln!("error: {}", e); }
        Ok(result) => { eprintln!("result: {:?}", result); }
    }

}
```

This fails to deserialize at runtime with the error "expected string for key `main.plugin`". I'm not sure why.

BTW there are workarounds, e.g. to use the untagged representation:

```rust
extern crate toml;
#[macro_use]
extern crate serde_derive;
extern crate serde;

#[derive(Serialize, Deserialize, Debug)]
pub struct Config {
    pub main: Option<Main>
}

#[derive(Serialize, Deserialize, Debug)]
#[serde(rename = "main")]
pub struct Main {
    pub plugin: Option<Plugin>,
}

#[derive(Serialize, Deserialize, Debug)]
#[serde(untagged)]
pub enum Plugin {
    Url { url: String },
    Bin { bin: String }
}

const SRC: &'static str = "
[main]
plugin = { \"bin\" = \"echo 'hello'\" }
";

fn main() {
    match toml::from_str::<Config>(SRC) {
        Err(e) => { eprintln!("error: {}", e); }
        Ok(result) => { eprintln!("result: {:?}", result); }
    }

}
```

azriel91:
[...]

Issue #222, opened 2018-01-10
How to deserialize toml config to a vector of enum struct varient?


HongbinZhou:
**TLDR:** How can I deserialize the following toml configuration str to Rust struct `Vec<Layer>`?

Deserialize config:

    [input]
    ndim = 2
    dims = [10, 100]

    [dense]
    outputs = 128

    [lstm]
    outputs =64
    return_seq = true

to a vector of `Layer`:

    [
       Input(ndim: 2, dims: [10, 100]),
       Dense(outputs: 128),
       Lstm(outputs: 64, return_seq: true)
    ]

while `Layer` is:

    enum Layer {
        Input { ndim: u32, dims: Vec<u32> },
        Lstm { outputs: u32, return_seq: bool },
        Dense { outputs: u32 },
    }

**My issue detailed description:**

I have a toml config file for a deep learning task similar as above shown, it
has one or more layers, each layer has its own key-value parameters. The layer
order matters.

I tried 3 ways but all failed.

**Try 1:** https://play.rust-lang.org/?gist=86b367e3148400f1090451c2fb1d7860&version=nightly

Using the following structures can deserilize the above config file but it
doesn't work for other configrations.

    #[derive(Serialize, Deserialize, Debug)]
    struct FixedLayersOnlyWorkForThisConfig {
        input: Input,
        dense: Dense,
        lstm: Lstm
    }
    #[derive(Serialize, Deserialize, Debug)]
    struct Layer {
        input: Input,
        lstm: Lstm,
    }
    #[derive(Serialize, Deserialize, Debug)]
    struct Input {
        ndim: u32,
        dims: Vec<u32>,
    }
    #[derive(Serialize, Deserialize, Debug)]
    struct Lstm {
        outputs: u32,
    }

For example it can't deserialize below config anymore:

    [input]
    ndim = 2
    dims = [10, 100]

    [dense]
    outputs = 128

**Try 2:** https://play.rust-lang.org/?gist=d88adf7a92c7063f955c59c85aa7149c&version=nightly

Using `toml::Value` `FromStr` can parse my config to toml internal structure:

    Table({
           "dense": Table({"outputs": Integer(128)}),
           "input": Table({"dims": Array([Integer(10), Integer(100)]), "ndim": Integer(2)}),
           "lstm" : Table({"outputs": Integer(64), "return_seq": Boolean(true)})
         })

But as toml used `BTreeMap` instead of `HashMap` for `Table`, **the order of layers
[...]

PR #264, opened 2018-10-09
Issue 255: Externally tagged enum deserialization

azriel91:
cc: #225, #222

~~The tuple variant hasn't been implemented yet, I'm having trouble figuring it out.~~
See #225 for ongoing discussion -- what should / should not be supported.

Short, the following TOML:

```toml
plain = "Plain"
plain_table = { Plain = {} }
tuple = { Tuple = { 0 = 123, 1 = true } }
struct = { Struct = { value = 123 } }
newtype = { NewType = "value" }
my_enum = [
    { Plain = {} },
    { Tuple = { 0 = 123, 1 = true } },
    { NewType = "value" },
    { Struct = { value = 123 } }
]
```

with the following types:

```rust
#[derive(Debug, Deserialize)]
struct Config {
    plain: MyEnum,
    plain_table: MyEnum,
    tuple: MyEnum,
    #[serde(rename = "struct")]
    structv: MyEnum,
    newtype: MyEnum,
    my_enum: Vec<MyEnum>,
}

#[derive(Debug, Deserialize)]
enum MyEnum {
    Plain,
    Tuple(i64, bool),
    NewType(String),
    Struct { value: i64 },
}
```

deserializes to:

```rust
Config {
    plain: Plain,
    plain_table: Plain,
    tuple: Tuple(
        123,
        true
    ),
    structv: Struct {
        value: 123
    },
    newtype: NewType(
        "value"
    ),
    my_enum: [
        Plain,
        Tuple(
            123,
            true
        ),
        NewType(
            "value"
        ),
        Struct {
            value: 123
        }
    ]
}
```

### DIFF
--- a/examples/enum_external.rs
+++ b/examples/enum_external.rs
@@ -11,7 +11,8 @@ extern crate serde_derive;
 #[derive(Debug, Deserialize)]
 struct Config {
     plain: MyEnum,
-    // tuple: MyEnum,
+    plain_table: MyEnum,
+    tuple: MyEnum,
     #[serde(rename = "struct")]
     structv: MyEnum,
     newtype: MyEnum,
@@ -29,12 +30,13 @@ enum MyEnum {
 fn main() {
     let toml_str = r#"
     plain = "Plain"
-    # tuple = { 0 = 123, 1 = true }
+    plain_table = { Plain = {} }
+    tuple = { Tuple = { 0 = 123, 1 = true } }
     struct = { Struct = { value = 123 } }
     newtype = { NewType = "value" }
     my_enum = [
         { Plain = {} },
-        # { Tuple = { 0 = 123, 1 = true } },
+        { Tuple = { 0 = 123, 1 = true } },
         { NewType = "value" },
         { Struct = { value = 123 } }
     ]"#;
--- a/examples/enum_external.rs
+++ b/examples/enum_external.rs
@@ -1,0 +1,40 @@
+//! An example showing off the usage of `Deserialize` to automatically decode
+//! TOML into a Rust `struct`, with enums.
+
+#![deny(warnings)]
+
+extern crate toml;
+#[macro_use]
+extern crate serde_derive;
+
+/// This is what we're going to decode into.
+#[derive(Debug, Deserialize)]
+struct Config {
+    plain: MyEnum,
+    // tuple: MyEnum,
+    #[serde(rename = "struct")]
+    structv: MyEnum,
+    my_enum: Vec<MyEnum>,
+}
+
+#[derive(Debug, Deserialize)]
+enum MyEnum {
+    Plain,
+    Tuple(i64, bool),
+    Struct { value: i64 },
+}
+
+fn main() {
+    let toml_str = r#"
+    plain = "Plain"
+    # tuple = { 0 = 123, 1 = true }
+    struct = { Struct = { value = 123 } }
+    my_enum = [
+        { Plain = {} },
+        # { Tuple = { 0 = 123, 1 = true } },
+        { Struct = { value = 123 } }
+    ]"#;
+
+    let decoded: Config = toml::from_str(toml_str).unwrap();
+    println!("{:#?}", decoded);
+}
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks for the PR! I don't personally have too many opinions about how we should map enums to toml, but I think we can probably pretty closely follow `serde_json` in this regard? Would it be possible to just clone whatever `serde_json` does for all these serializations?

azriel91:
Finally got round to this, `serde_json` serializes the value like this:

<details>
<summary>json serialized</summary>

```json
{
    "plain": "Plain",
    "plain_table": "Plain",
    "tuple": {
        "Tuple": [
            -1,
            false
        ]
    },
    "struct": {
        "Struct": {
            "value": -1
        }
    },
    "newtype": {
        "NewType": "abc"
    },
    "my_enum": [
        "Plain",
        {
            "Tuple": [
                -1,
                false
            ]
        },
        {
            "NewType": "abc"
        },
        {
            "Struct": {
                "value": -1
            }
        }
    ]
}
```
</details>

The structure is similar, the differences:

* JSON array allows you to have mixed types, so
    - Tuples are represented using arrays in JSON. (mixed types are invalid in TOML, so we need to use a TOML table)
    - Array of enum variants is valid in JSON: `["Plain", { "Tuple": [ -1, false ] }]`
* In TOML it would be ergonomic to represent plain enum variants as `"String"`, but represented as a TOML table when used in an array

I'll work on getting serialization to work that way, since it's the closest we can get to JSON

azriel91:
I've gotten deserialization matching the above (and tested! yay). For serialization, the tuple variant currently emits a toml array, which is invalid if you have a tuple variant with different types.

I'm not sure whether it's okay to break backwards compatibility and emit a table -- should I:

* **A:** Try and be clever and emit an array if the tuple fields are the same type, and a table if they are different types?
* **B:** Make it always emit an inline table

azriel91:
I'll push serialization to a separate branch &ndash; I haven't managed to get it *correct*, let alone clean. In case you're happy with having deserialization without serialization as a partial state.

alexcrichton:
Oh dear sorry about that @Aaronepower! I'm somewhat low on reviewing bandwidth for this at the moment, so some assistance in reviewing this would be greatly appreciated!

Aaronepower:
@alexcrichton Sure no problem!

@azriel91 There seem to be a lot of unrelated style changes to the codebase. (Adding semicolons, changes the formatting of arguments, extra commas, etc.) Are you perhaps using rustfmt? I think removing those tidy ups would make it easier to see the changes. For example there's 940 changes in `src/de.rs`, but as far as I can tell there's actually 144 lines that aren't tidying up the code.

azriel91:
Yeap, the first commit is purely rustfmt, so the relevant diff would be the ones after that.
I can push over this without that if you prefer

Aaronepower:
@azriel91 If you could that would be very helpful. I've found that trying to review through individual commits is a lot of work as you have to keep track of the code from previous commits as well new code.

azriel91:
There we go

Aaronepower:
@azriel91 Generally this looks good to me. Can you add some tests that should always fail with this kind of deserialisation?

azriel91:
"Yes!", but first I have to understand what you mean 😅. There's these possible meanings:

* array of tables: unsupported but potentially supported later (implement as "pending")
* invalid variant / fields in variants (should panic, useful error message)
* something else?

azriel91:
I've pushed tests + changes for the first two bullet points

Aaronepower:
@azriel91 It seems that last commit has caused some errors in the CI.

azriel91:
ah yeap, only noticed it yesterday, pushed a quick fix 🤞

Aaronepower:
@alexcrichton Have a look yourself, but it looks good to me.

alexcrichton:
Er sorry for the delay!

I unfortunately still don't really understand most of this code, the serde implementation has always been opaque at best to me. I'll trust y'all though!

PR #267, opened 2018-10-22
[Need Help] External Enum Serialization

azriel91:
*See https://github.com/alexcrichton/toml-rs/pull/264 for deserialization*

Serialization half of #225. I haven't managed to get it functioning, let alone clean. Learnings so far:

* `src/ser.rs` contains the `Serializer` which stores some information to track "what type is used in the last output array"
* When serializing hierarchical data (e.g. array containing an inline table), it should instantiate a separate `Serializer` (or store the context in some sort of stack)
* It doesn't do that, so when serializing part of the inline table's data into a toml string, and it checks "is my last array type also a string", it fails because it stores "inline_table" as the last type in the array.

Fixing this would probably require restructuring how the serialization context is stored. I'm taking a break from working on this so if anyone would like to continue it, please do!

---

On the branch, there's an `tests/enum_external_serialize.rs` file which is meant to show how it will be used. Also, probably want to implement pretty print and non-pretty print properly

### DIFF
--- a/examples/enum_external.rs
+++ b/examples/enum_external.rs
@@ -11,7 +11,8 @@ extern crate serde_derive;
 #[derive(Debug, Deserialize)]
 struct Config {
     plain: MyEnum,
-    // tuple: MyEnum,
+    plain_table: MyEnum,
+    tuple: MyEnum,
     #[serde(rename = "struct")]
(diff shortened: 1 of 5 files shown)


epage:

Issue #270, opened 2018-10-28
"map key was not a string" for newtype struct


spease:
```
#[derive(Serialize)]
pub struct MyType(String);
```

Seems to be the cause of the error. Even though ultimately the newtype should get serialized as a string, it still triggers an error. I'm guessing the problem is in
```impl ser::Serializer for StringExtractor```
as even the serialize_newtype functions immediately return an error as opposed to simply calling the serialize method for the wrapped value.

ehuss:
This appears to be fixed, please reopen if I am misinterpreting.

Issue #196, opened 2017-07-05
Unable to deserialize newtype structs


Jascha-N:
Test case:
```rust
#[derive(Deserialize, Serialize)]
struct A {
    b: B
}

#[derive(Deserialize, Serialize)]
struct B(u32);

fn test() {
    let encoded = toml::to_string(&A { b: B(2) }).unwrap();
    toml::from_str::<A>(&encoded).unwrap();
}
```

Results in:
```
Error { inner: ErrorInner { kind: Custom, line: None, col: 0, message: "invalid type: integer `2`, expected tuple struct B", key: ["b"] } }
```

Serialization of newtype structs seems to work fine, but deserialization fails. I'm pretty sure it is expected that serializing followed by deserializing should (almost?) always work, but in any case `toml-rs` should be able to deserialize newtype structs.

alexcrichton:
Thanks for the report!

Jascha-N:
Thanks, but unfortunately I seemed to have reduced the test case too much. There are still issues. 

New test case:
```rust
#[derive(Debug, Deserialize, Serialize)]
struct A {
    b: B
}

#[derive(Debug, Deserialize, Serialize)]
struct B(Option<C>);

#[derive(Debug, Deserialize, Serialize)]
struct C {
    x: u32,
    y: u32,
    z: u32
}

fn main() {
    let input = A { b: B(Some(C { x: 0, y: 1, z: 1 })) };
    println!("{:?}", input);
    let encoded = toml::to_string(&input).unwrap();
    println!("{}", encoded);
    let output = toml::from_str::<A>(&encoded).unwrap();
    println!("{:?}", output);
}
```

Results in:

```
Error { inner: ErrorInner { kind: Custom, line: None, col: 0, message: "invalid type: map, expected tuple struct B", key: ["b"] } }
```

And replacing `input` with `A { b: B(None) }` leads to:

```
Error { inner: ErrorInner { kind: Custom, line: None, col: 0, message: "missing field `b`", key: [] } }
```

`serde_json` seems to handle these cases fine, but I'd rather not switch to JSON for my configuration files.

alexcrichton:
Ah good points. I believe the first one (with `Some`) is handled with https://github.com/alexcrichton/toml-rs/commit/f6673dfdb3f93bacfbf0c0cab4d78bc4154eec0c although the latter with `None` I'm not sure where it's going awry...

PR #271, opened 2018-10-28
Allow newtype key

spease:
See #270

### DIFF
--- a/src/ser.rs
+++ b/src/ser.rs
@@ -1420,11 +1420,11 @@ impl ser::Serializer for StringExtractor {
         Err(Error::KeyNotString)
     }
 
-    fn serialize_newtype_struct<T: ?Sized>(self, _name: &'static str, _value: &T)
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

Er but actually now that I've merged it, mind adding a test for this as well?

spease:
Sorry I haven't replied. I've got the notice marked as unread in my inbox to remind me. That being said, I noticed that deserialization gets hung up on `invalid type: map, expected tuple struct Map`. Looks like it's already ticketed as #196 but it was believed to be solved.

Excerpts of the problem code:
```
/// A path of Identifiers
#[derive(Debug, Display, Eq, Hash, PartialEq, Serialize)]
pub struct Path(String);
```
```
#[derive(Debug, Deserialize, Serialize)]
pub struct Map(HashMap<Path, MapEntry>);
```

Not sure when I'll get to this as it's for a spare-time group project so it's lower priority atm.

alexcrichton:
Ok no worries, thanks again!

PR #272, opened 2018-10-28
Support tuple Serde types for `Value`

hcpl:
With https://github.com/alexcrichton/toml-rs/pull/245 already allowing tuples, tuple structs and tuple variants for string serialization, there is no reason `toml::Value` should not have the same capability as well!


alexcrichton:
Thanks!

alexcrichton:
Thanks!

PR #275, opened 2018-11-21
Fix trailing space after date.

### LINKED ISSUE #273: Allow space after YYYY-MM-DD dates
The following TOML is currently invalid:

```toml
date = 2018-10-10 
```
Pretty hard to see the issue while looking at it. There is a space after the last digit, causing the parser to return an error as it couldn't parse a datetime.

I had a look at other TOML parsers and some parse it as a date while others error. The grammar seems to indicate that no space should be allowed but that's a bit hard to spot for a human (https://github.com/getzola/zola/issues/525).
If the current behaviour is not going to change, can we get a different error so we can tell the user what to look for?

ehuss:
The space between date and time was being eagerly skipped when it shouldn't.

Fixes #273.


alexcrichton:
Thanks!

PR #242, opened 2018-05-14
Add the preserve_order feature

Keats:
Fix #232 

This is pretty much a copy/paste of the serde_json preserve_order feature.
I'm not entirely sure how to test it with the whole `test-suite` subcrate though, would it be ok to add a few tests in the main crate instead?

Issue #266, opened 2018-10-21
Support for u128, i128 integer types 


AregevDev:
Pretty self-explained, serde is already support them, will be nice to see them in the TOML format

alexcrichton:
Thanks for the report! Can you gist an example of what doesn't work to have a good test case to check in?

AregevDev:
Sure thing, will be done when I will be back

AregevDev:
https://gist.github.com/AregevDev/79dcde42ee06c23d3f5a892596001667
There you go!

AregevDev:
Any news?

AregevDev:
I serialized a struct with two 128 bit integers successfully but deserializing them is the tricky part

AregevDev:
Built from source and modified a little

AregevDev:
Solved, I can deserialize i/u128 which are not overflowing the i/u64::max_value() perspectively

nullchinchilla:
I'd like to bump this issue a bit, since it's quite a critical feature for my use case.

TheDarkula:
I opened a pull request to add this functionality: #445

epage:
Maintenance of this crate has moved to the https://github.com/toml-rs/toml repo.  As a heads up, we plan to move `toml` to be on top of `toml_edit`, see https://github.com/toml-rs/toml/issues/340.

Closing this out.  If this is still a problem, feel free to recreate this issue in the new repo.

PR #278, opened 2018-12-28
Add the preserve_order feature (rebased version)

Kerollmops:
Here is the #242 PR from @Keats rebased on master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,9 @@ before_script:
   - pip install 'travis-cargo<0.2' --user && export PATH=$HOME/.local/bin:$PATH
 script:
   - cargo test
+  - cargo test --features preserve_order
   - cargo test --manifest-path test-suite/Cargo.toml
+  - cargo test --manifest-path test-suite/Cargo.toml --features preserve_order
   - rustdoc --test README.md -L target
   - test "$TRAVIS_RUST_VERSION" != "1.15.0" && cargo doc --no-deps || echo "skipping cargo doc"
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ script:
   - cargo test
   - cargo test --features preserve_order
   - cargo test --manifest-path test-suite/Cargo.toml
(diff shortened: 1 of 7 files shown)


alexcrichton:
Thanks!

@ehuss you mentioned that you're working on a toml parser preserving style, I'm curious if it'd be at all related to something like this? Were you thinking that'd be a totally standalone project?

(mostly looking to minimize major version bumps)

ehuss:
🤔 I'm not sure, I've been meaning to talk to you about that.  It could go either way.  Right now I'm developing it in tree.  If it is a separate crate, toml-rs will need to expose a few low-level methods and data structures which doesn't sound too appealing.  However, it's a fairly large amount of code and I don't want to just dump it on the project.  There's at least a few weeks of work left to do, so maybe once it's more together we could take a look and discuss the options?

As for the relation to this, I think code-wise it is independent.  I'm not really touching the serde or Value parts.

There's maybe a bigger question of sharing code with serde_json.  For example, to really support #266 it might be nice to use the `Number` type from serde_json.  There's a lot of good stuff in serde_json, and just copying it seems a little weird.

alexcrichton:
Ah ok, do you know if the changes you'd need, if it goes in this repository, will require breaking changes? I'm fine evaluating the best location for the work once it's done!

I'm somewhat wary about `Number` and such as integers in toml are apparently defined as signed 64-bit integers, so while we should handle 128-bit integers I don't think we can handle values outside the range of a 64-bit signed integer.

ehuss:
The only changes I see so far is adding some new `pub` items.  In particular:
- `Deserializer::line` — this is the main parsing function that will return everything including whitespace and comments.
- `Line` enum (returned by `line`)
- The "raw" key and value types from Deserializer that will contain whitespace/comments.
- `Deserializer::value` and `Deserializer::dotted_key` for parsing just a key or value.

Right now I have them `pub(crate)` which would unfortunately require a minimum bump to 1.18.   Otherwise I don't see any other fundamental changes.

Good point on the integers.

PR #280, opened 2019-01-07
Fix disallowing duplicate table headers

### LINKED ISSUE #279: Duplicate table not detected.
The following TOML is accepted, but it should be rejected due to the duplicate table.

```toml
[dependencies.openssl-sys]
version = "0.5.2"

[dependencies]
libc = "0.1"

[dependencies]
bitflags = "0.1.1"
```

Discovered in these 3 `Cargo.toml` manifests on crates.io:
- https://crates.io/crates/pnetlink/0.0.1
- https://crates.io/crates/pnetlink/0.0.2
- https://crates.io/crates/openssl/0.5.2

I don't think I understand the `cur_parent` dance in `MapVisitor` well enough to fix this.

alexcrichton:
This commit fixes #279 where a case of duplicate table headers slipped
through the cracks. This also adds an option to disable this new
validation to allow Cargo to preserve backwards compatibility.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,9 @@ facilitate deserializing and serializing Rust structures.
 """
 categories = ["config", "encoding", "parser-implementations"]
 
+[workspace]
+members = ['test-suite']
+
 [badges]
 travis-ci = { repository = "alexcrichton/toml-rs" }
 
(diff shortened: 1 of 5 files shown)


ehuss:
Nice!  I like the workspace, too, was wondering why that hadn't been done earlier.

PR #289, opened 2019-02-16
Chrono integration.

### LINKED ISSUE #159: Add time/chrono to access date/time values
Currently, date/time values are only represented by internal structures that only support conversion to/from strings. Adding the ability to convert these types to structures from the time and/or chrono crate would enable people to deal wth these types much more nicely.

cc https://github.com/Rustaceans/rust-cologne/issues/29

quadrupleslap:
This isn't a breaking change, and it fixes #159.

Concerns:

- This is still  wrong because it doesn't treat the date types as separate types so, for example, it allows both a date and a date-time in the same array.
- Chrono's subsecond output doesn't truncate zeros (it always pads to a length that's a multiple of three) so I changed some test cases.
- The date parser still has no error messages.
- Chrono capitalizes the `T` in `DateTime` (and so does the TOML spec), but **toml-rs** doesn't.


ehuss:
Can this be done so that the dependency on chrono is optional?

quadrupleslap:
Probably? The bigger problem is that

> it's pretty obvious that they are different types (and hence cannot be mixed in arrays), since they are listed as different types in the spec.

so apparently the rules when de/serializing arrays and stuff also needs to be changed.

ehuss:
Thinking about this more... Is the goal to be able to natively handle deserializing to types like `struct S { ndt: chrono:NaiveDateTime }`?  If that's the goal, would it be possible to go the simple route of adding something like this to `ValueDeserializer`:

```rust
    fn deserialize_str<V>(self, visitor: V) -> Result<V::Value, Self::Error>
    where
        V: de::Visitor<'de>,
    {
        match self.value.e {
            E::Datetime(s) => visitor.visit_str(s),
            _ => self.deserialize_any(visitor),
        }
    }
```

Chrono's serde support should pick that up.

Or are you more concerned about accessing chrono types from `toml::Value`?

I'm not too concerned about the homogeneous array case.  With the above approach, if you have the type `Vec<chrono::NaiveTime>`, then chrono will handle rejecting mixed time types.

quadrupleslap:
Yeah, `toml::Value` isthe problem. Also, what about the serialization side? chrono would just serialize it to a string.

epage:

PR #294, opened 2019-03-20
Fix parsing of repeated delimiters inside multi-line strings

tmiasko:
The previous implementation of `read_string`, when looking for
delimiters ending the multi-line string and failing to find exactly
three in succession, always put a single delimiter back. This is
incorrect when exactly two delimiters are present.

Put back the correct number of delimiters depending on how many have
been already consumed from input.

The issue could be triggered only with `Owned` representation of
`MaybeString`, since otherwise `push` operation is no-op.


alexcrichton:
Oops, thanks!

PR #295, opened 2019-04-02
Fix warnings for deprecated usages of trim_{left,right}_matches.

ErichDonGubler:


ehuss:
Thanks!  This raises the MSRV to 1.30, though Alex says he's fine with "current stable-1".

ErichDonGubler:
Huh? So far as I can tell, this crate is already MSRV 1.30 because [the `$crate` prefix is being used](https://doc.rust-lang.org/edition-guide/rust-2018/macros/macro-changes.html#macros-with-crate-prefix). Am I missing something?

ehuss:
For this one, trim_end/trim_start weren't stabilized until 1.30.

ErichDonGubler:
Understood. Thanks!

PR #296, opened 2019-04-02
Make importing `toml_internal` macro unnecessary when using the `toml` macro

ErichDonGubler:


ehuss:
Thanks!  This raises the MSRV to 1.30, though Alex says he's fine with "current stable-1".

ErichDonGubler:
Huh? So far as I can tell, this crate is already MSRV 1.30 because [the `$crate` prefix](https://doc.rust-lang.org/edition-guide/rust-2018/macros/macro-changes.html#macros-with-crate-prefix) is [already being used](https://github.com/alexcrichton/toml-rs/blob/d729bf9c53fcfe8f1e5506a938b88d81526b55a4/src/macros.rs#L37). Am I missing something?

ehuss:
The issue is that `$crate::toml_internal!` was not supported until 1.30.  That is specifically, a macro with a multi-segment path.  `$crate` prefix for regular items has always been supported.

ErichDonGubler:
Ahhh, I see. You're right, then. Thanks!

PR #313, opened 2019-07-28
Show "column" in Error messages

zertosh:
Error messages now show "column N" when the line number is available.
Not only does this make the messages more useful, but it's inline with how
`serde_json` displays them: https://github.com/serde-rs/json/blob/e3a2c3af66f4/src/error.rs#L368

Subsequent commits will add `line` and `col` to additional error kinds.


alexcrichton:
Sounds reasonable to me, thanks!

PR #314, opened 2019-07-30
Add line and column to all Errors

zertosh:
This change adds offset information to Errors as they bubble up the visitor
chain, and before exiting, adds line and column information based on that
offset. The effect is that _most_ Errors now have correct location information,
and in other cases, it's close (but not quite as exact as it should be).

The new test (`de-errors.rs`), serves to show the positioning of custom
deserializer Errors. These are the most problematic ones, because table-level
errors, require guessing the position from the elements. Unfortunately,
this isn't great, but it's better than nothing. The tests highlight where
the positioning falls short.

This is the continuation of #312 and #313.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -29,7 +29,7 @@ where
 {
     match str::from_utf8(bytes) {
         Ok(s) => from_str(s),
-        Err(e) => Err(Error::custom(e.to_string())),
+        Err(e) => Err(Error::custom(None, e.to_string())),
     }
 }
 
@@ -87,6 +87,7 @@ struct ErrorInner {
     kind: ErrorKind,
     line: Option<usize>,
     col: usize,
+    at: Option<usize>,
     message: String,
     key: Vec<String>,
 }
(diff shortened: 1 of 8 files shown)


src/de.rs
```diff
@@ -1710,10 +1764,8 @@ impl<'a> Deserializer<'a> {
     }
 
     fn error(&self, at: usize, kind: ErrorKind) -> Error {
-        let mut err = Error::from_kind(kind);
-        let (line, col) = self.to_linecol(at);
-        err.inner.line = Some(line);
-        err.inner.col = col;
+        let mut err = Error::from_kind(Some(at), kind);
+        err.fix_linecol(|at| self.to_linecol(at));
```
alexcrichton: Since `at` is passed in above, isn't this not necessary because it'll never be run?

zertosh: Here, the `at` passed to `Error::from_kind` and the `at` argument in the `fix_linecol` closure, are def the same. But you still need to call `fix_linecol`. Do you mean that you prefer:

```
let mut err = Error::from_kind(Some(at), kind);
err.fix_linecol(|_| self.to_linecol(at));
```

Happy to PR is.

zertosh:
No worries! Can I bug you for a new release with these changes? Thanks!

alexcrichton:
Sure thing, done!

src/de.rs
alexcrichton: Oh nah I think I misunderstood what was happening here, seems fine as-is!

PR #315, opened 2019-08-02
Allow conversion from de::Error to std::io::error

jwdeitch:
Hi there! 

Use case is to allow something like this:

```
pub fn load(file_path: &str) -> Result<TaskToml, std::io::Error> {
    let t: TaskToml = toml::from_str(
        &fs::read_to_string(file_path)?
    )?;

    Ok(t)
}
```


alexcrichton:
Thanks!

PR #318, opened 2019-08-13
Derive Clone+PartialEq for Spanned

JustAPerson:
Should any other traits be derived?


alexcrichton:
sounds reasonable to me!

PR #319, opened 2019-08-13
Simplified Logic when comparing discriminants

NickHackman:
Removed matches that were used to compare Enum variants for equality in
favor of using `discriminant` from [1.21.0](https://doc.rust-lang.org/std/mem/fn.discriminant.html)
that's cleaner and more idiomatic.


alexcrichton:
👍

PR #320, opened 2019-08-13
Fix error line/column when using CRLF line endings.

ehuss:
The new exposure to line/col in errors uncovered an issue with how `to_linecol` was computing the line/col with a CRLF document. It was not including the CR characters in the offset computation, causing the result to skew for each line.


alexcrichton:
Nice! Wanna merge when ci is green and publish?

PR #321, opened 2019-08-19
Fix clippy warnings and errors

NickHackman:
Hi,

Trying to resolve issue #139 and hit a roadblock when it comes to how to handle certain situations. A majority of the warnings and errors have been resolved. I have 1 remaining warning and a question revolving around the use of `drop` in `src/ser.rs`.

Firstly, the one remaining warning

```
warning: methods called `to_*` usually take self by reference; consider choosing a less ambiguous name
   --> src/tokens.rs:505:17
    |
505 |     fn to_owned(&mut self, input: &str) {
    |                 ^^^^^^^^^
    |
    = note: `#[warn(clippy::wrong_self_convention)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#wrong_self_convention
```
Personally, I think that the name of the method properly describes the purpose. Just to entertain it what would be a better name?

## Drop

Drop is used frequently in `src/ser.rs` on `Results`, which caused errors for Clippy, but I'm not sure how to properly handle that. Right now, I have a plausible solution, but that just ignores the result. Should the error properly result in a warning message? How do you think this should be handled?

Thanks in advance I really appreciate it :smile:

### DIFF
--- a/src/ser.rs
+++ b/src/ser.rs
@@ -25,6 +25,8 @@
 //! # fn main() {}
 //! ```
 
+#![allow(unused_must_use)]
+
 use std::cell::Cell;
 use std::error;
 use std::fmt::{self, Write};
@@ -422,7 +424,7 @@ impl<'a> Serializer<'a> {
 
     fn display<T: fmt::Display>(&mut self, t: T, type_: &'static str) -> Result<(), Error> {
         self.emit_key(type_)?;
-        drop(write!(self.dst, "{}", t));
+        write!(self.dst, "{}", t);
         if let State::Table { .. } = self.state {
             self.dst.push_str("\n");
         }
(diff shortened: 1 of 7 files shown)


src/ser.rs
```diff
@@ -425,7 +424,7 @@ impl<'a> Serializer<'a> {
 
     fn display<T: fmt::Display>(&mut self, t: T, type_: &'static str) -> Result<(), Error> {
         self.emit_key(type_)?;
-        drop(write!(self.dst, "{}", t));
+        write!(self.dst, "{}", t);
```
alexcrichton: I think this should probably be `?` or remain as `drop` perhaps?

alexcrichton: (same for all other instances of this)

NickHackman: I think the `?` approach is more idiomatic, errors are now being mapped to `ser::Error` via `ser::Error::custom`

If all is well, I'm going to open this as an actual PR

alexcrichton:
Thanks!

PR #324, opened 2019-08-25
Make Map impls generic

est31:
Preparatory work for #236 .

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -6,7 +6,7 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-//! A map of String to toml::Value.
+//! A map of a key to a value.
 //!
 //! By default the map is backed by a [`BTreeMap`]. Enable the `preserve_order`
 //! feature of toml-rs to use [`LinkedHashMap`] instead.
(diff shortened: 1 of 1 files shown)


est31:
@alexcrichton the type is public but I don't think this is a breaking change because of the orphan rule. As they are foreign, downstream code can't implement those traits themselves, even for `Map<TypeProvidedByDownstream, sth else>`.

As for the rationale: my goal is to add spans support to `toml::Value`. I'm driven by a personal use for it but ultimately even Cargo would benefit from this as it could now print proper toml snippets with the error when the user provided an erroneous Cargo.toml. I've described my approach in this comment https://github.com/alexcrichton/toml-rs/issues/236#issuecomment-524202343 . Unlike this PR, the larger change does involve a breaking change but I'm not sure one can be easily avoided without duplicating the code.

Part of the approach would be to replace the `pub type Table = Map<String, Value>;` typedef with a `pub type Table = Map<Spanned<String>, Spanned<Value>>;` typedef. Currently the relevant impls are all on the hardcoded `Map<String, Value>` though, so I thought I'd make them generic while I'm at it. This is what this PR is a part of. Technically, it's not needed as one could simply change all occurences of `Map<String, Value>` to `Map<Spanned<String>, Spanned<Value>>` (or to just use the typedef from value.rs), but making it generic feels like it's more idiomatic to me. If I get your ok, I'll continue making all Map code generic. But I can also just change it to `Table` instead.

est31:
Basically there are two questions:

* Should Map be generic or not?

* Which approach should be used to bring spans to users of `Value`? I see three options here: either duplicating the types manually as suggested by @udoprog , or making `Value` generic so that it can either store spans or not store them (basically this would be automatic duplication), or just always storing spans (my favourite). The first two options would benefit from a generic map, the third one doesn't technically need it but it might still be a great idea.

est31:
re-r? @alexcrichton

est31:
Note that this PR does constitute a breaking change now as some of the exposed types (iterators, entry, etc) are generic now.

est31:
Rebased.

est31:
I've rebased this PR now, but I guess it can only be added to toml-rs's next breaking semver release. In general it's not a large priority for the `toml-spanned-value` crate as I can work around it by just copying the file. Of course it would be nicer if upstream had support, so if this got merged some day in the future it would be really cool!

epage:

PR #327, opened 2019-08-30
Add a spanned value 

### LINKED ISSUE #95: Get the position of a Value in the file
It would be nice for error reporting to be able to get the position (the span) of a specific TOML value in the file. 

For example, I am parsing a configuration file for my project, and the user entered a string where an integer is needed. I would like to report the position of the error in addition with the usual `"this key must be an integer"` message.

I do not know how hard this would be to implement, and how the API could look like. Having something like

``` rust
struct Span {
    file: Path,
    line_start: usize,
    line_end: usize,
    col_start: usize,
    col_end: usize,
}

enum Value {
    String(String, Span),
    Integer(i64, Span),
    Float(f64, Span),
    Boolean(bool, Span),
    Datetime(String, Span),
    Array(Array, Span),
    Table(Table, Span),
}
```

would be rather inconvenient to use. Maybe something like that, with a duplication of methods on `Value` and `Entry`.

``` rust
struct Entry {
    value: Value,
    span: Span
}
```

If you think this is not too hard, I can give it a try.

### LINKED ISSUE #236: Support emitting spans when deserializing structures

est31:
Fixes #236.
Fixes #95.

Builds on #324, #328 and #333.

Limitations:

* No ability to get spans of dotted tables as well as arrays defined with `[[key]]` (inline tables as well as [] arrays do work). While we could probably special case tables/arrays that are contiguous, in the general case the toml data format allows settings where tables/arrays are spread over parts of the file. In these cases, we set both start and end of the span to 0, as this is easily detectable and no interesting value would ever have such a span. User code needs to be able to handle these cases.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -558,6 +558,36 @@ impl<'de, 'b> de::Deserializer<'de> for MapVisitor<'de, 'b> {
         visitor.visit_newtype_struct(self)
     }
 
+    fn deserialize_struct<V>(
+        self,
+        name: &'static str,
+        fields: &'static [&'static str],
+        visitor: V,
+    ) -> Result<V::Value, Error>
+    where
+        V: de::Visitor<'de>,
+    {
+        if name == spanned::NAME && fields == [spanned::START, spanned::END, spanned::VALUE] {
+            // TODO we can't actually emit spans here for the *entire* table/array
+            // due to the format that toml uses. Setting the start and end to 0 is
+            // *detectable* (and no reasonable span would look like that),
+            // it would be better to expose this in the API via proper
+            // ADTs like Option<T>.
+            let start = 0;
+            let end = 0;
+
+            let res = visitor.visit_map(SpannedDeserializer {
+                phantom_data: PhantomData,
+                start: Some(start),
+                value: Some(self),
+                end: Some(end),
+            });
+            return res;
+        }
+
+        self.deserialize_any(visitor)
+    }
+
     fn deserialize_enum<V>(
         self,
         _name: &'static str,
@@ -591,7 +621,7 @@ impl<'de, 'b> de::Deserializer<'de> for MapVisitor<'de, 'b> {
 
     serde::forward_to_deserialize_any! {
         bool u8 u16 u32 u64 i8 i16 i32 i64 f32 f64 char str string seq
-        bytes byte_buf map struct unit identifier
+        bytes byte_buf map unit identifier
         ignored_any unit_struct tuple_struct tuple
     }
 }
@@ -850,6 +880,14 @@ impl<'de> de::Deserializer<'de> for ValueDeserializer<'de> {
         bool u8 u16 u32 u64 i8 i16 i32 i64 f32 f64 char str string seq
         bytes byte_buf map unit identifier
         ignored_any unit_struct tuple_struct tuple
+    }
+}
+
+impl<'de, 'b> de::IntoDeserializer<'de, Error> for MapVisitor<'de, 'b> {
+    type Deserializer = MapVisitor<'de, 'b>;
+
+    fn into_deserializer(self) -> Self::Deserializer {
+        self
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -150,6 +150,7 @@
 
 pub mod map;
 pub mod value;
+pub mod spanned_value;
 #[doc(no_inline)]
 pub use crate::value::Value;
 mod datetime;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -149,8 +149,10 @@
 #![warn(rust_2018_idioms)]
 
 pub mod map;
-pub mod value;
 pub mod spanned_value;
+pub mod value;
+#[doc(no_inline)]
+pub use crate::spanned_value::SpannedValue;
 #[doc(no_inline)]
 pub use crate::value::Value;
 mod datetime;
--- a/src/map.rs
+++ b/src/map.rs
@@ -6,7 +6,7 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-//! A map of String to toml::Value.
+//! A map of a key to a value.
 //!
 //! By default the map is backed by a [`BTreeMap`]. Enable the `preserve_order`
 //! feature of toml-rs to use [`LinkedHashMap`] instead.
@@ -14,12 +14,12 @@
 //! [`BTreeMap`]: https://doc.rust-lang.org/std/collections/struct.BTreeMap.html
 //! [`LinkedHashMap`]: https://docs.rs/linked-hash-map/*/linked_hash_map/struct.LinkedHashMap.html
 
-use crate::value::Value;
 use serde::{de, ser};
 use std::borrow::Borrow;
 use std::fmt::{self, Debug};
 use std::hash::Hash;
 use std::iter::FromIterator;
+use std::marker::PhantomData;
 use std::ops;
 
 #[cfg(not(feature = "preserve_order"))]
@@ -29,7 +29,8 @@ use std::collections::{btree_map, BTreeMap};
 use linked_hash_map::{self, LinkedHashMap};
 
 /// Represents a JSON key/value type.
-pub struct Map<K, V> {
+#[derive(Clone, PartialEq)]
+pub struct Map<K: Ord + Hash, V> {
     map: MapImpl<K, V>,
 }
 
@@ -38,7 +39,7 @@ type MapImpl<K, V> = BTreeMap<K, V>;
 #[cfg(feature = "preserve_order")]
 type MapImpl<K, V> = LinkedHashMap<K, V>;
 
-impl Map<String, Value> {
+impl<K: Ord + Hash, V> Map<K, V> {
     /// Makes a new empty Map.
     #[inline]
     pub fn new() -> Self {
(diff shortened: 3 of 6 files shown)


est31:
@alexcrichton this is ready for a first review. I've added tests that show the capabilities of the new API.

Also, I wonder about your suggestions for how to expose table keys in the API. The basic problem is that you can't just store Map<Span<String>, Span<ValueKind>>, because users should be able to index that map with a string only. So you'd have to do something like Map<String, MapEntry> where MapEntry contains the span of both the table index and the Span<ValueKind>. This however causes issues with the index implementation of ValueKind due to the return type being inconsistent between index types (arrays return Span<ValueKind> only). I'd be okay with shipping SpannedValue first without support for key spans and only figuring it out later.

est31:
> This looks like it complicates the Value API 

Changing the Value API to contain spans was my old suggestion that was [already rejected](https://github.com/alexcrichton/toml-rs/pull/324#issuecomment-525790430) by @alexcrichton . Instead I added a completely new `SpannedValue` like he suggested.

>  The design of Spanned is such that if you don't use it then it doesn't affect you

This is already what this PR is doing. `Value` is kept completely untouched. In fact the SpannedValue introduction itself isn't even a breaking change. There are breaking changes needed, but only for some iterators and other "unimportant" data structures on the Map type in #324.

> relatively tiny amount of people would use.

Beyond uses in my own projects, my hope is that one day cargo will use it to display proper errors with toml snippets when you do something wrong in `Cargo.toml`. It would also be very cool if `cargo-udeps` could point to the unused crate in the toml directly. Right now I have two projects which would benefit from this PR. This isn't random. Especially in Rust, toml is very often used as a configuration format for humans to edit, instead of just an opaque data format handled only by machines. Judging from the threads of #236 and #95, I don't seem to be the only one with this request.

dtolnay:
Keep in mind that a huge number of crates depend on toml (751 as of today) so even multiple reports is not necessarily significant, and doing all the things that people want is not a long term successful strategy.

I guess the changes that got pulled in here from #324 are what I am objecting to. I disagree that these are "unimportant", because they make the crate harder to use. For example if I am browsing the docs of Entry and want to or_insert a value -- what type do I need to pass? In practice always `Value`, but the docs would only say `V`. In figuring this stuff out out, a new user is forced to pay the complexity of SpannedValue existing when they would never use it.

For the crates that need a SpannedValue, having it provided by an external crate with its own Entry and Iterator types etc would be just as good.

For cargo-udeps would deserializing the dependencies to `BTreeMap<Spanned<String>, Value>` be sufficient?

est31:
> I disagree that these are "unimportant"

I should probably have mentioned what I've meant with "unimportant": most user's code won't be broken by it because they don't store these types in their structures. Most people never mention the type anywhere. Compared to a hypothetical breaking change of `Value` to `Spanned<Value>`, the breakage is minimal.

> they make the crate harder to use

Those APIs are constructed after the generic APIs in libstd and liballoc. Names, functionalities, etc are all the same. Most rustaceans should be familiar with those APIs already, and that PR brings them closer. In fact, for me it was quite surprising that these functions/types weren't generic already. The code right now is in a (for me at least) confusing hybrid state of some things being generic (`Map` itself is `Map<K,V>`), while all relevant impls are only on `Map<String, Value>`. I understand why `Map` was originally changed to be non-generic: fundamentally, generics can be confusing to people and so your point isn't completely wrong, and monomorphization causes code bloat, slow compile times, etc. But in this case, the complexity cost already seems to be paid.

est31:
Rebased due to #328 and #329.

est31:
@dtolnay as you're already here, how would you expose the span of the keys? The issue with `Map<Spanned<String>, Value>` is that you can't index by a string any more. Also I might need help in getting the information through serde. This question needs an answer whether the span functionality is implemented in this crate or a separate one.

dtolnay:
On indexing, here is one possibility:

```rust
use ref_cast::RefCast;
use std::borrow::Borrow;

mod any_spanned {
    use ref_cast::RefCast;

    #[derive(RefCast, PartialEq, Eq, PartialOrd, Ord)]
    #[repr(transparent)]
    pub struct AnySpanned<T: ?Sized>(T);
}

pub use self::any_spanned::*;

#[allow(non_snake_case)]
pub fn AnySpanned<T: ?Sized>(value: &T) -> &AnySpanned<T> {
    AnySpanned::ref_cast(value)
}

impl<T, Q: ?Sized> Borrow<AnySpanned<Q>> for Spanned<T>
where
    T: Borrow<Q>,
{
    fn borrow(&self) -> &AnySpanned<Q> {
        AnySpanned(self.value.borrow())
    }
}
```

Now we index by wrapping the key to indicate that any span is fine:

```rust
pub fn demo(map: BTreeMap<Spanned<String>, ()>) {
    map[AnySpanned("...")]
}
```

On getting the information through Serde, where does it fail currently? Serde is generally fine with non-string keys, so ideally the same scheme that works for values would also work for keys.

est31:
Thanks @dtolnay this is a quite interesting method! One issue is probably that the map index method is just a wrapper over [get](https://doc.rust-lang.org/stable/std/collections/struct.BTreeMap.html#method.get) which for `BtreeMap` at least requires that "the ordering on the borrowed form must match the ordering on the key type". There are less docs for `LinkedHashMap`, but I assume that similar holds true for it as well (it uses the `Hash` trait instead). So we'll need to implement custom `Ord` and `Hash` impls for `Spanned<T>` that ignore the start and end fields.

est31:
Also, according to [Eq docs](https://doc.rust-lang.org/stable/std/cmp/trait.Ord.html#how-can-i-implement-ord):

> Implementations of PartialEq, PartialOrd, and Ord must agree with each other. That is, a.cmp(b) == Ordering::Equal if and only if a == b and Some(a.cmp(b)) == a.partial_cmp(b) for all a and b. 

So I guess we'll have to change `Spanned<T>` to ignore its span innards for all these traits.

est31:
Hmm so in the context of #332 and the general problem that such an AnySpanned only works with unsafe or requires two additional crates (that add compile time), I think we can just use Spanned with dummy data like start and end both set to zero. As outlined above, Spanned has to ignore the span data anyway for its `Hash` and `{Partial,}{Eq,Ord}` implementations anyway and we'd avoid having to pull in two crates (ref_cast would add two new crates to our dep graph) or using unsafe manually.

alexcrichton:
I think I agree seeing this now with @dtolnay that a `SpannedValue` is probably overkill, but with https://github.com/alexcrichton/toml-rs/pull/333 is this still necessary for your use case?

est31:
> with #333 is this still necessary for your use case?

With #328 and #333 merged, most of the `de.rs` changes needed for `SpannedValue` are in master already. That's really cool.

As for putting `SpannedValue` into a separate crate, there are some challenges. First, the PR uses multiple `pub(crate)` items:

* `DatetimeFromString` from datetime.rs
* `datetime::FIELD` from datetime.rs
* `spanned::{START,END,VALUE}` from spanned.rs
* `SpannedVisitor` from spanned.rs (previously private to the module, now exposed as pub(crate))

Furthermore, it's doing some changes to the crate:

* `Hash`/`Ord`/`Eq`/etc. impls of `Spanned<T>` are made to ignore the span values and only care about the `T` (need is outlined in the thread) 
* `de.rs` is getting "support" for deserializing dotted tables and arrays by setting their span end and start to 0. If this change is not present, deserializing any toml file with dotted tables into a `SpannedValue` would fail. The data inside those dotted tables can contain span data again (and the *keys* of those tables have spans too), but toml's data model allows for cases where a table can be spread over the entire file. 

The `de.rs` change, if welcome, can be put into its own PR, and I guess it is *possible* to copy-paste the `pub(crate)` items as well as all of `Spanned<T>`, adding the needed  `Hash`/`Ord`/`Eq`/etc. impls. So you'd end up being able to share lots of code between the crates.

But the API between the crates would involve very private things like `spanned::{NAME,START,END,VALUE}` and `datetime::FIELD` which have values are like `$__toml_private_Spanned`, or the order in which the spanned fields are serialized. If the API is exposed, I guess breakage risk can be minimized.  E.g. documenting the serialization order as well as making the string constants part of the public API (with the caveat that the value might change between versions of the toml crate).

In general, I'd prefer if `SpannedValue` could live in upstream, but as you said you don't want this, I'll file a separate `de.rs` PR as well as PRs to document the order as well as making the constants public. Will try to do it this week. In the case you change your mind, I invite you to reopen this PR ;).

est31:
@dtolnay check out the `DateTimeOrTable` struct in `spanned_value.rs`. It's needed to either deserialize a spanned key or a `DateTime` key. As SpannedValue can contain both, it needs to detect this from the data. The same already exists in value.rs but here it only distinguishes between a string key and a `DateTime` key. You are the serde expert here :) so if you know of a replacement, please tell me!

Also, I'm not sure whether @alexcrichton wants the `Spanned<T>` changes to ignore the spans for the eq, ord, hash, etc impls, or whether the separate crate is expected to provide its very own `Spanned<T>`. Right now the presence of SpannedValue in toml-rs provides good justification.

alexcrichton:
I don't personally have much of a preference of whether `Spanned<T>`'s derived trait impls ignore or take into account the actual spans, if a compelling argument can be made one way or another I'm fine either way.

est31:
List of PRs:

* Span "support" for dotted tables: #340
* Exposing the constants and documenting order: TODO

est31:
Crate's at: https://github.com/est31/toml-spanned-value

PR #332, opened 2019-09-08
Forbid unsafe code

Shnatsel:
`#![forbid(unsafe_code)]` annotation makes rustc abort compilation if there are any unsafe blocks in the crate.

Presence of this annotation is picked up by tools such as [cargo-geiger](https://github.com/anderejd/cargo-geiger) and lets them ensure that there is indeed no unsafe code as opposed to something they couldn't detect (e.g. unsafe added via macro expansion, etc).

This also adds a badge advertising that unsafe code is forbidden. Being 100% safe is a desirable property for a crate and should be advertised. The badge currently links to Secure Code WG's community outreach repo, but I'm fine with changing the link or making the badge non-clickable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 [![Latest Version](https://img.shields.io/crates/v/toml.svg)](https://crates.io/crates/toml)
 [![Documentation](https://docs.rs/toml/badge.svg)](https://docs.rs/toml)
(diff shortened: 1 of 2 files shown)


alexcrichton:
Seems reasonable to me! I'd prefer to not update the README now though. Also mind dropping a comment near `#![forbid]` to explain why it's there?

Shnatsel:
@alexcrichton done. Sorry for the delay.

alexcrichton:
👍

PR #333, opened 2019-09-09
Support deserializing spanned keys

est31:
This adds support for deserialization of table key span information. cc #327


alexcrichton:
Looks good to me, thanks!

PR #337, opened 2019-09-22
Correctly handle table headers backline prepending

GuillaumeGomez:


GuillaumeGomez:
rustfmt is definitely my arch enemy...

src/ser.rs
```diff
@@ -691,7 +691,7 @@ impl<'a> Serializer<'a> {
 
         match *state {
             State::Table { first, .. } => {
-                if !first.get() {
+                if !first.get() || !self.dst.is_empty() {
```
ehuss: Just a few questions, I'm not terribly familiar with the serializer side of the library (I'm having a bit of trouble understanding what `first` actually means here).

Should `State::Array` also have a similar check?

Is it even necessary to check if `first` is set?  When would `dst` be empty and `first` be false?

GuillaumeGomez: I tried to make as little changes as possible so I kept the `first` check. I can always try out without it if you want?

> Should `State::Array` also have a similar check?

I don't think we want it for arrays (at least I think so?). The point of this change is to handle this case:

```toml
[foo]
x = 1

[foo.something]
y = 2
```

In the current code, there wouldn't have the empty line between the two whereas it should. I think that an array shouldn't have the empty line, no?

ehuss: The reason I ask about arrays is because the following:

```rust
assert_eq!(
    Table(map! {
        "test" => Table(map!{"val" => Integer(2)}),
         "z" => Table(map!{
            "foo" => Array(vec![Table(map!{"hello" => Integer(3)})])
         })
    })
    .to_string(),
    "[test]\n\
    val = 2\n\
    [[z.foo]]\n\
    hello = 3\n"
);
```

Notice there is no empty line between the `[test]` table and the `[[z.foo]]` array.  This also seems to depend on the alphabetic ordering (`[[a.foo]]` does include a blank line).

We can wait to see if Alex has any opinion on how it should work, since it does seem a little wonky.

alexcrichton: Oh dear it's been years at this point since I read this code. I agree with @ehuss though that arrays/tables need to be treated the same here, and I don't think checking `self.dst` is right because you could also be serializing into a non-empty string to start with or something like that, but we should produce the same serialization regardless

GuillaumeGomez:
Ok, updated so the check is better. :)

I also added the same for arrays.

alexcrichton:
Hm while this does technically work it's basically the same code. There's already a `first` field which should be handling this and I'm not sure why it's not working. I think it'd be best to investigate why that's not working as-is and see if there's a bug around its handling

GuillaumeGomez:
The `first` field value is "correct" since we are in the first table. However, if we add a table inside another table, then maybe we should change this `first` value? Well, I can do it if you want like this.

GuillaumeGomez:
@alexcrichton I can't find any easy solution to this problem. I might very well miss something obvious but if you try to change `first` value in `SerializerSeq` when initializing it, it has unwanted side effects. You also can't change the `first` value when running because it'd alter the parent too since it's a cell, something you don't want either. Therefore, all changes I tried didn't work and started to get bigger and bigger. This time, I think I reached the best possible change without side effects.

However, if you find a way to do it without changing half the code or having side effects, I'd be more than happy to have it!

PS: I updated the tests because one was useless and indent was weird sometimes...

GuillaumeGomez:
Any news?

GuillaumeGomez:
Anyone? I'd really like to have this merged and published so we can finally fix our issue in gtk-rs... :-/

GuillaumeGomez:
Any news?

alexcrichton:
Sorry I don't have the time to invest in this (clearly!). I don't feel this is the right solution and I'm not comfortable merging this personally. I don't have the time though to help develop the right answer for this issue, but the general shape of it would have to do with tweaking the emission rather than after-the-fact trying to pop off newlines.

epage:

PR #339, opened 2019-10-04
ValueAfterTable fix

PSeitz:


PSeitz:
Not sure if this is possible, but would be nice if the conversion toml->json->toml would be the same as the input.

I noticed the conversion back, e.g. yields an empty object.

```toml
[[kanji]]

[kana]
ent_seq = '1000000'

```

```json
{
  "kanji": [
    {}
  ],
  "kana": {
    "ent_seq": "1000000"
  }
}
```

alexcrichton:
Hey sorry for the late review!

Looks like CI is failing here, but I also agree that we should probably fix that input. I'm not 100% sure that this is the right fix though, can you try seeing what the output looks like with tables before/after the array output here? (also a mixtures of values before/after as well)

vpzomtrrfrt:
This fails with empty objects:
```rust
fn main() {
    let value = serde_json::json!({"objects": []});
    let as_str = toml::to_string(&value).unwrap();
    let reparsed: serde_json::Value = toml::from_str(&as_str).unwrap();

    println!("{}", as_str);
    println!("{}", serde_json::to_string(&reparsed).unwrap());
}
```

Serializes as:
```
[[objects]]
```

Which is not the same, since it represents an array containing an empty object:
```
{"objects":[{}]}
```

tami5:
🤔 Is this issue fixed in another pr?

At first I though hash map is not supported, so I started using Vec but that too yield same error. it works okay as long as the field does not contain any value

```rust
    pub queries: HashMap<String, Request>,
    pub queries: Vec<(String, Request)>,
```

epage:

PR #340, opened 2019-10-05
Support for dotted table spans

est31:
This provides "support" for dotted table spans by setting them to 0. While special cases could get special case logic, in the general case toml allows for dotted tables without well defined spans.

In this document, the span for the `dependencies` table is for example not well-defined:

```toml
[dependencies.serde]
optional = true

[package]
name = "toml-spanned-value"

[dependencies]
toml = "0.7"
```

The package table for example does have a well-defined span and thus could be special-cased.


alexcrichton:
Looks reasonable to me, sorry for the delay in getting to this. Thanks though!

PR #341, opened 2019-10-05
Expose Datetime and Spanned constants

est31:
See https://github.com/alexcrichton/toml-rs/pull/327#issuecomment-531988931 and the related discussion.

### DIFF
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -1,3 +1,5 @@
+//! Custom Datetime structures
+
 use std::error;
 use std::fmt;
 use std::str::{self, FromStr};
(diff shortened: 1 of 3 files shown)


alexcrichton:
There was a comment [below yours](https://github.com/alexcrichton/toml-rs/pull/327#pullrequestreview-288956011) which would entail not exporting these, and would it be possible to explore that? I would prefer to not export these low-level details which is something we can't change easily

est31:
The code that these constants are needed for is:

* The [DatetimeOrTable](https://github.com/est31/toml-spanned-value/blob/ec985b5dc15634834214973afd2310f870c6f5ea/src/spanned_value.rs#L550-L595) struct that helps differentiating between DateTime and a Spanned struct.

* The custom [Spanned](https://github.com/est31/toml-spanned-value/blob/ec985b5dc15634834214973afd2310f870c6f5ea/src/spanned.rs) struct that I'm using so that I can set Eq, Hash, etc. impls in a custom way.

Elimination of the second case seems possible as long as you want to merge the PR to change the impls. For the first case, I don't know. In the thread you linked I've asked @dtolnay to help me and in fact that's why I waited so long until I filed this PR. `DatetimeOrTable` is some kind of hybrid deserializer that either deserializes `DateTime` or `Spanned<Table>`, depending on the data present. While this crate has such a struct as well, the struct in this crate is differentiating between `DateTime` and `Table` without the `Spanned`. So exposing this crate's `DateTimeorTable` wouldn't work. How would calling the `Deserialize` impl work as @dtolnay proposes?

est31:
So I figured out a way. It's not beautiful. In fact, it's quite ugly. And it also makes assumptions about how the Spanned is being deserialized (as a map in the visitor).

est31:
I'm working on a patch for the second case.

est31:
Ok I've opened #344 . Either this PR or #344 is needed for the `toml-spanned-value` crate. #340 is needed in any case. #324 would be cool to be merged, but isn't required, as `toml-spanned-value` has its own copy/fork of map.rs that has the PR applied.

est31:
Superseded by #344

Issue #310, opened 2019-07-20
Support tabs in strings.


ehuss:
The spec was recently updated to allow bare tabs in basic strings: https://github.com/toml-lang/toml/pull/627

PR #343, opened 2019-10-09
feat: support tabs in basic strings

jkcclemens:
See #310.

I don't think this needs anything else? Please correct me if I'm wrong!


alexcrichton:
Thanks!

PR #344, opened 2019-10-13
 Spanned: impl PartialEq, Eq, Hash, PartialOrd, Ord in terms of the value

est31:
This is because we want to be able to index into HashMap<Spanned<String>, T>
with a dummy span and get results where only the content has to match.

For more info, see #327 and #341


alexcrichton:
Sorry for the delay in getting to this!

est31:
Thanks for merging!

PR #349, opened 2019-10-25
Decrease deserialization complexity from quadratic to linear

### LINKED ISSUE #342: Large TOML document performance
I'm attempting to read a large-ish TOML document like so:

```rust
println!("Reading toml size {}", s.len());
let t = toml::from_str(s).unwrap();
```

where the `str` size is around 4MB (this is an auto-generated TOML, obviously). The file is essentially a lot of very small tables in the following format:

```toml
[table-id-1]
name = "Foo"
url = "http://example.com"
tags = ["tag-name"]
```

The loading time is unreasonable, spinning the CPU up to 100% and taking way too long, over a minute before I kill the process.

Am I wrong to expect this library to be able to handle files that big?

est31:
Fixes #342.

In particular, see my comment https://github.com/alexcrichton/toml-rs/issues/342#issuecomment-546190394

Values that I recorded on my machine for running the code `measure_time(n, |i| format!("[header_no_{}]\n", i))` function for varying `n`:

| benchmark  | optimizations | before this PR | after this PR |
|------------|---------------|----------------|---------------|
|          1k|             no|           170ms|           48ms|
|         10k|             no|         14 191ms|          483ms|
|        100k|             no|             n/a|         5 077ms|
|          1k|            yes|             5ms|            4ms|
|         10k|            yes|           311ms|           39ms|
|        100k|            yes|         63 850ms|          364ms|

You can nicely see how before this PR, a 10x increase in data meant a 100x increase in time spent, while after the PR it only means a 10x increase.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -323,6 +323,16 @@ impl<'de, 'b> de::Deserializer<'de> for &'b mut Deserializer<'de> {
     }
 }
 
+// Builds a datastructure that allows for efficient sublinear lookups.
+// The returned HashMap contains a mapping from table header (like [a.b.c])
+// to list of tables with that precise name. The tables are being identified
+// by their index in the passed slice. We use a list as the implementation
+// uses this data structure for arrays as well as tables,
+// so if any top level [[name]] array contains multiple entries,
+// there are multiple entires in the list.
+// The lookup is performed in the `SeqAccess` implementation of `MapVisitor`.
+// The lists are ordered, which we exploit in the search code by using
+// bisection.
 fn build_table_indices<'de>(tables: &[Table<'de>]) -> HashMap<Vec<Cow<'de, str>>, Vec<usize>> {
     let mut res = HashMap::new();
     for (i, table) in tables.iter().enumerate() {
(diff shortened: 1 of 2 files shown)


src/de.rs
```diff
@@ -364,20 +393,27 @@ impl<'de, 'b> de::MapAccess<'de> for MapVisitor<'de, 'b> {
...
-                            Some(header) => headers_equal(&header, &prefix),
-                            None => false,
+                    .map(|v| v.1.clone())
+                    .collect::<Vec<_>>();
+                self.table_pindices
+                    .get(&prefix_stripped)
+                    .and_then(|entries| {
+                        let start = entries
+                            .binary_search(&self.cur)
+                            .unwrap_or_else(std::convert::identity);
```
alexcrichton: I'd personally prefer if this were `unwrap_or_else(|i| i)`

alexcrichton:
👍

alexcrichton:
It looks like the tests added here [may be causing a spurious failure on CI](https://github.com/alexcrichton/toml-rs/runs/277767964#step:6:132)?

est31:
Huh, that's weird. It works on my machine but it failed on CI previously as well. As a result I increased the tolerance value, but it seems further increases are needed. The difference is quite large. I can maybe increase the sample size, which should lead to less noise. And maybe increase the multiplier as well to allow for even larger tolerances. Does that sound reasonable?

alexcrichton:
CI machines (VMs) tend to be extremely noisy in terms of measurements, so I think it's fine to perhaps remove the test and move it to a benchmark which can be manually tracked over time. This is pretty unlikely to regress.

PR #351, opened 2019-10-28
Replace the test added by #349 with a bench

est31:
CI environments can be noisy and while the test worked great
locally on my machine, it didn't on the CI environment.
This replaces the test with a (manually tracked) benchmark.
As per https://github.com/alexcrichton/toml-rs/pull/349#issuecomment-546998173


alexcrichton:
Thanks!

Issue #357, opened 2019-11-06
Support mixed-type arrays.


ehuss:
The spec has been updated to allow arrays of mixed types (https://github.com/toml-lang/toml/pull/676), so it might be good to relax that restriction.

ehuss:
Looks like this is fixed by #358.

PR #358, opened 2019-11-10
Mixed type arrays

daubaris:
Regarding #357, reduced restrictions for homogeneous arrays. Added support for mixed types inside arrays.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -11,7 +11,6 @@ use std::f64;
 use std::fmt;
 use std::iter;
 use std::marker::PhantomData;
-use std::mem::discriminant;
 use std::str;
 use std::vec;
 
@@ -146,10 +145,6 @@ enum ErrorKind {
         /// Actually found token type
         found: &'static str,
     },
-
-    /// An array was decoded but the types inside of it were mixed, which is
-    /// disallowed by TOML.
-    MixedArrayType,
 
     /// A duplicate table definition was found.
     DuplicateTable(String),
@@ -1827,13 +1822,7 @@ impl<'a> Deserializer<'a> {
             if let Some(span) = self.eat_spanned(Token::RightBracket)? {
                 return Ok((span, ret));
             }
-            let at = self.tokens.current();
             let value = self.value()?;
-            if let Some(last) = ret.last() {
-                if !value.same_type(last) {
-                    return Err(self.error(at, ErrorKind::MixedArrayType));
-                }
-            }
             ret.push(value);
             intermediate(self)?;
             if !self.eat(Token::Comma)? {
@@ -2118,7 +2107,6 @@ impl fmt::Display for Error {
             }
             ErrorKind::NumberInvalid => "invalid number".fmt(f)?,
             ErrorKind::DateInvalid => "invalid date".fmt(f)?,
-            ErrorKind::MixedArrayType => "mixed types in an array".fmt(f)?,
             ErrorKind::DuplicateTable(ref s) => {
                 write!(f, "redefinition of table `{}`", s)?;
             }
@@ -2180,7 +2168,6 @@ impl error::Error for Error {
             ErrorKind::Wanted { .. } => "expected a token but found another",
             ErrorKind::NumberInvalid => "invalid number",
             ErrorKind::DateInvalid => "invalid date",
-            ErrorKind::MixedArrayType => "mixed types in an array",
             ErrorKind::DuplicateTable(_) => "duplicate table",
             ErrorKind::RedefineAsArray => "table redefined as array",
             ErrorKind::EmptyTableKey => "empty table key found",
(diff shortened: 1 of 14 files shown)


src/de.rs
```diff
@@ -147,10 +146,6 @@ enum ErrorKind {
         found: &'static str,
     },
 
-    /// An array was decoded but the types inside of it were mixed, which is
-    /// disallowed by TOML.
-    MixedArrayType,
```
alexcrichton: Could this be left in temporarily to avoid a breaking change, but we can file an issue to delete it on the next breaking change?

src/ser.rs
```diff
@@ -496,11 +492,7 @@ impl<'a> Serializer<'a> {
             State::Array { type_, .. } => type_,
             _ => return Ok(()),
         };
-        if let Some(prev) = prev.get() {
-            if prev != type_ {
-                return Err(Error::ArrayMixedType);
-            }
-        } else {
+        if let None = prev.get() {
             prev.set(Some(type_));
```
alexcrichton: This may no longer be necessary?

src/de.rs
daubaris: Yeah, you're right. I'm on it.

src/ser.rs
daubaris: Without it some tests (e.g., `simple_show` and `table` in `tests/display.rs`) fail.

src/de.rs
daubaris: @alexcrichton Also, what could be the best approach to handle `dead_code` warning, since `MixedArrayType` isn't used anywhere. I don't think it is a good idea to put `#[allow(dead_code)]`, but I can't come up with another solution..

dtolnay: This enum is private so it isn't breaking to remove variants.

src/ser.rs
```diff
@@ -124,10 +124,6 @@ pub enum Error {
     #[doc(hidden)]
     KeyNewline,
 
-    /// Arrays in TOML must have a homogenous type, but a heterogeneous array
-    /// was emitted.
-    ArrayMixedType,
```
alexcrichton: Oh my mistake, this is the public variant which would be great to stick around for now

src/ser.rs
alexcrichton: Hm can you gist the failures? Do you know why this is still needed? If so, can you explain what this field is doing?

src/de.rs
alexcrichton: Er my mistake, I didn't realize this was private. Yes it's private so it's fine to delete.

src/ser.rs
daubaris:
```
---- simple_show stdout ----
thread 'simple_show' panicked at 'assertion failed: `(left == right)`
  left: `"[1, 2[]"`,
 right: `"[1, 2]"`', test-suite/tests/display.rs:20:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.

---- table stdout ----
thread 'table' panicked at 'Unable to represent value as string: ValueAfterTable', src/libcore/result.rs:1165:5


failures:
    simple_show
    table
```

alexcrichton: Sorry I don't have the time to dig into this right now, but can you try to learn why the tests are failing? We shouldn't have to record type information any more and this may just be related to something else. I doubt that full type information needs to continue to be recorded and it's probably just something like "was anything seen"

daubaris: Sure, I'll look into this :+1:

daubaris: @alexcrichton Yeah so the problem is at: https://github.com/alexcrichton/toml-rs/blob/75129040cd041b380f4f1d8f2685928392aa9f55/src/ser.rs#L989-L1013

It jumps to `None` case and adds additional `[]`. Since without `prev.set(Some(type_));` we don't set a type.

alexcrichton: Could this be adjusted to a custom enum then? It's basically three states of "not started", "started", and "started as a table"

daubaris: So does that mean that we should basically eliminate the regular types (`integer`, `float`, etc.) in the `enum State` by changing the `&'static str` with a custom `enum` of your described "not started", "started", "started as a table" states? I assume that because we really shouldn't care about the array type at this point since we allow the array to contain multiple types.

alexcrichton: That sounds good to me!

daubaris: @alexcrichton I've made some changes. I think that we don't even need the `NotStarted` state, since the type is an `Option` within a `Cell`. I might be wrong though. Anyways open for discussion.

alexcrichton:
👍 

Thanks!

samuelcolvin:
Any chance of a release to include this?

alexcrichton:
Sure thing, done!

samuelcolvin:
thanks so much!

PR #362, opened 2019-12-03
[enum] improve parsing for dotted tables (#225)

johansglock:
Use the current referenced table instead of assuming we should only have
1 table, allowing for more complex parsing of enum structures.
Similarly, depending on the context, values can be in either the table
or in the current values context. Take them as appropriate.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -11,6 +11,7 @@ use std::f64;
 use std::fmt;
 use std::iter;
 use std::marker::PhantomData;
+use std::ops::Deref;
 use std::str;
 use std::vec;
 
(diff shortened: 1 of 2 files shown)


alexcrichton:
Sorry for the delay in reviewing this, but I unfortunately do not have time to review PRs to this crate right now. I hope to get a chance to review this in the future but I'm likely to be busy for some time now.

tjkirch:
This would be very nice to have.

I've fixed the (trivial) conflicts, here's an updated patch: https://gist.github.com/tjkirch/fbbd40fe5966ddad52a1012590de0e3f

I could also do a new PR if that's easier.

I can confirm this does fix the problem for me - I'm able to deserialize an externally tagged enum inside a dotted table, which fails without the patch.

skomski:
It would be great if this could land sometime because inline tables are a bit verbose.

danobi:
Bump :)

Would be great to get this in.

brianc118:
Just bumping this again - this would be super useful

ehuss:
Can someone provide a summary of what the problem is, what this tries to fix, how it goes about fixing it, why are there several FIXMEs, etc?  It is difficult to page in the context around this, and there isn't much information to go on here.

Issue #360, opened 2019-11-15
Publicly expose de::ErrorInner


Luthaf:
I would like to customize the error message on deserialization failure, and it looks like most of the data is already there in `de::ErrorInner`, but the struct is private

https://github.com/alexcrichton/toml-rs/blob/75129040cd041b380f4f1d8f2685928392aa9f55/src/de.rs#L93-L100

Would you be OK with exposing `ErrorInner` publicly, either through a `Deref` implementation or by forwarding some getters to access its fields?  `line` and `col` are already accessible through `de::Error::line_col()`.

If you agree, I am happy to make a PR for this.

alexcrichton:
Thanks for the report! This is an internal error though and would need design for actually being exposed, so I would prefer to not just mark this `pub` and reexport it. If you'd like to add a new constructor though and/or setters/getters, seems fine to add those in a PR!

PR #364, opened 2019-12-10
Expose part of de::ErrorInner publicly

Luthaf:
See #360. 

The second commit exposes `de::ErrorKind` publicly, if this is not desired I can remove it. It should be fine to expose since it is already marked as `non_exhaustive`, using `#[doc(hidden)]`.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -98,9 +98,9 @@ struct ErrorInner {
     key: Vec<String>,
 }
 
-/// Errors that can occur when deserializing a type.
+/// Kinds of errors that can occur when deserializing a type.
 #[derive(Debug, PartialEq, Eq, Clone)]
-enum ErrorKind {
+pub enum ErrorKind {
     /// EOF was reached when looking for a value
     UnexpectedEof,
 
@@ -2023,6 +2023,11 @@ impl Error {
     /// Get the keys associated with this error
     pub fn key(&self) -> &[String] {
         &self.inner.key
+    }
+
+    /// Get the error kind of this error
+    pub fn kind(&self) -> &ErrorKind {
+        &self.inner.kind
     }
 
     fn from_kind(at: Option<usize>, kind: ErrorKind) -> Error {
--- a/src/de.rs
+++ b/src/de.rs
@@ -94,7 +94,6 @@ struct ErrorInner {
     line: Option<usize>,
     col: usize,
     at: Option<usize>,
-    message: String,
     key: Vec<String>,
 }
 
@@ -160,7 +159,7 @@ pub enum ErrorKind {
 
     /// A custom error which could be generated when deserializing a particular
     /// type.
-    Custom,
+    Custom(String),
 
     /// A tuple with a certain number of elements was expected but something
     /// else was found.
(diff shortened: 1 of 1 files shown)


alexcrichton:
Could you expand a bit on the rationale for this change? What are you looking for to get out of these new accessors?

Luthaf:
My use case is to customize the error message displayed to the user, adding specific context and rewording them.

I am using TOML as input file for my code, and currently manually transforming from `toml::Value` to the structs. I would like to use serde instead, but the current error messages are not ideal.

alexcrichton:
Sorry but I unfortunately do not have time to review PRs to this crate right now. I hope to get a chance to review this in the future but I'm likely to be busy for some time now.

Luthaf:
No problem, this is not urgent on my side =)

Luthaf:
@alexcrichton, do you foresee having some time to review this in the near future? Or is there another maintainer for this crate?

No pressure, I would like to know before I look for other solutions 😃

src/de.rs
```diff
@@ -2015,6 +2015,21 @@ impl Error {
         self.inner.line.map(|line| (line, self.inner.col))
     }
 
+    /// Get the message associated with this error
+    pub fn message(&self) -> &str {
+        &self.inner.message
+    }
```
ehuss, line 2021: This seems a little off, because this only returns the message for the custom kind.  I'm kinda wondering why the message isn't part of `ErrorKind::Custom`.

src/de.rs
```diff
@@ -2015,6 +2015,21 @@ impl Error {
...
+
+    /// Get the keys associated with this error
+    pub fn key(&self) -> &[String] {
+        &self.inner.key
+    }
```
ehuss, line 2026: This method also seems a little off, because it can return an empty slice in many situations.  I'm almost thinking it would be cleaner to return an option?

src/de.rs
Luthaf: You mean using the `message` name for this function is weird? I am happy to use `custom_message` instead. 

Moving the message to `ErrorKind::Custom` also works, although it would be a slightly larger change.

src/de.rs
Luthaf: Sure if you prefer it this way 😃 . I don't see the issue with an empty slice though. I would expect `Some(&[])` to have a different meaning from `None` when dealing with `Option<&[T]>`

src/de.rs
ehuss: Yea, ideally the message would be embedded in `ErrorKind::Custom(String)`, and remove this method.  It just seems a little too odd to have a method like this (and a dedicated `message` field) for one kind.

src/de.rs
ehuss: Yea, I think it would be better to be more explicit about the semantics here.

Luthaf:
@alexcrichton or @ehuss, I would appreciate if one of you could give this a new look =) Thanks!

smoelius:
Thank you, @Luthaf.

@ehuss Would you still consider adding this functionality?

smoelius:
@Luthaf In the event this PR is not merged, may I ask, did you find an alternative solution? My use case is the same as yours:
> My use case is to customize the error message displayed to the user, adding specific context and rewording them.

Luthaf:
I don't have a need for this anymore, but thanks a lot for picking up this project!

PR #368, opened 2020-01-15
Fix a couple errors in DateStrEmitter

mgsloan:
Before this change, there were a couple cases in `DateStrEmitter` report `KeyNotString` errors instead of `DateInvalid`.


alexcrichton:
Thanks!

PR #369, opened 2020-01-15
Make UnsupportedType error more informative

mgsloan:
`UnsupportedType` with no further info is not very helpful when trying to serialize a complex structure.  This adds more information so that the source of the problem is clearer to the developer.

I realize that this is a breaking change.  An alternative I considered was adding a variant to the enum, but I think this is better.  Cases where users are explicitly matching on this probably ought to be updated - hopefully there are few in the wild.

### DIFF
--- a/src/ser.rs
+++ b/src/ser.rs
@@ -114,13 +114,20 @@ pub enum Error {
     ///
     /// Currently the TOML format does not support serializing types such as
     /// enums, tuples and tuple structs.
-    UnsupportedType,
+    UnsupportedType {
+        /// Name of the type for which serialization is not supported, if any.
+        name: Option<String>,
+        /// Name of the variant for which serialization is not supported, if any.
+        variant: Option<String>,
+        /// The variety of type which failed to serialize .
+        variety: String,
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks for the PR! This looks good to me, but yeah we'll want to hold off on merging since this is a breaking change. Seems fine to me though to land in the future!

mgsloan:
You're welcome, thanks for reviewing it!

chipsenkbeil:
@alexcrichton, was there any interest in merging this? Or is there an alternative way to handling this? I'm getting hit by this error and have no idea where in my complex type this would be happening.

Was using serde in combination with toml to both read from a config file (working) and produce an example config file (failing).

alexcrichton:
Indeed yes, still interested in merging, but holding off  until this crate has a breaking release, which it currently isn't scheduled yet to have.

mgsloan:
Cool!  Please do let me know when I should rebase it for that purpose, or
feel free to copy-modify the change.

epage:

PR #373, opened 2020-01-31
Allow empty table keys

mgsloan:
I found this issue by writing a roundtrip property test using quickcheck - https://github.com/alexcrichton/toml-rs/compare/master...mgsloan:roundtrip-property-test .

Quoting https://github.com/toml-lang/toml/tree/68076ffc6d9a150d48f1df964551283542ad47bc#user-content-keys

> A bare key must be non-empty, but an empty quoted key is allowed (though discouraged).

See also the discussion in https://github.com/toml-lang/toml/issues/432


mgsloan:
Looks like empty keys were quite intentionally disallowed.  I can see this being diagnostically nice for many applications, but toml allows it, so I think it's better to match the spec.

mgsloan:
Note that the reason this was found by a property test is that the serializer will happily output invalid toml when serializing a table with an empty key.

joelthelion:
Could this fix be merged? The current behavior clearly goes against the spec.

ehuss (approved):
Thank for the PR!  It is a little strange that they decided to allow empty keys, but since it is explicitly noted in the spec, I agree it should probably be allowed.

PR #376, opened 2020-02-15
Implementd Deserializer for &Value.

Kixunil:
This adds efficient implementation of Deserializer for borrowed Value.
It allows deserializing from Value without cloning, as it borrows the
data from the deserializer (original value).

I'm unsure about tests. The change is pretty straightforward, but it might be worth testing. How should I approach it? Should I add tests? If yes, should I just test if the values deserialized from `&Value` equal to those deserialized from `Value` or something more?


Kixunil:
Ping?

epage:

PR #381, opened 2020-03-11
Implement writers from to serde_json

### LINKED ISSUE #216: Why can't you encode to any "Write"?

bbqsrc:
Resolves #216. Alternative to #257.

The interface for the writers is precisely the same as in `serde_json` except using the `toml` Error type. As this codebase is set to not allow `unsafe`, I changed the unsafe utf8 conversions to their lossy equivalent, though we guarantee UTF-8 so this should be truly equivalent and lossless.

Didn't do readers yet as they're a bit more involved. Also fixed a couple of typos while I was there.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -161,7 +161,9 @@ mod datetime;
 
 pub mod ser;
 #[doc(inline)]
-pub use crate::ser::{to_string, to_string_pretty, to_vec, to_vec_pretty, to_writer, to_writer_pretty, Serializer};
+pub use crate::ser::{
+    to_string, to_string_pretty, to_vec, to_vec_pretty, to_writer, to_writer_pretty, Serializer,
+};
 pub mod de;
 #[doc(inline)]
 pub use crate::de::{from_slice, from_str, Deserializer};
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -160,10 +160,10 @@ pub use crate::value::Value;
 mod datetime;
 
 pub mod ser;
-#[doc(no_inline)]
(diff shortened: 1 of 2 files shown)


bbqsrc:
Ah, I see `cargo test` doesn't run all the tests. Good stuff. I'll fix these.

alexcrichton:
Thanks for the PR, but I do not want to add a type parameter to all serialization/deserialization in this crate. That causes compilation to slow down as it causes quite a bit of monomorphization in the extreme case. I would prefer that everything was just always done with in-memory strings since TOML isn't really meant for gigabyte-big things anyway.

bbqsrc:
If I changed it to `dyn Write` would it be acceptable?

alexcrichton:
I would personally say it's probably still not worth it, wrappers if necessary could just be conveniences over the string-based version. I'm not sure what the performance impact of switching to `dyn Write` would be.

Mr-Andersen:
> Thanks for the PR, but I do not want to add a type parameter to all serialization/deserialization in this crate. That causes compilation to slow down as it causes quite a bit of monomorphization in the extreme case. I would prefer that everything was just always done with in-memory strings since TOML isn't really meant for gigabyte-big things anyway.

So, no `from_reader`?

PR #385, opened 2020-03-22
Improve error message when parsing unquoted string

### LINKED ISSUE #379: Generate better error for invalid value syntax.
An invalid value can often provide a confusing error "invalid number".  For example:

```text
remote = origin
```

One common way this can happen is forgetting the quotes for a string. This can happen if someone is familiar with INI syntax (or syntax like git's config files).

The root of the problem is that the parser assumes any "keylike" thing is a number [here](https://github.com/alexcrichton/toml-rs/blob/7a43ac9e9c28afec8414f6d924024f5d33fe8fa2/src/de.rs#L1431).  I think it would be good to check that the string has a form that starts with a number-like thing, and if it doesn't, generate a better error that explains that it expected a valid TOML value.  Valid starting characters are 0-9, `-`, `inf`, and `nan` (`+` is handled as a separate token).

aleksator:
Fixes #379

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -1462,9 +1462,7 @@ impl<'a> Deserializer<'a> {
     }
 
     fn parse_keylike(&mut self, at: usize, span: Span, key: &'a str) -> Result<Value<'a>, Error> {
-        let lowercase = key.to_ascii_lowercase();
-        let lowercase = lowercase.as_str();
-        if lowercase == "inf" || lowercase == "nan" {
+        if key == "inf" || key == "nan" {
             return self.number_or_date(span, key);
         }
 
--- a/src/de.rs
+++ b/src/de.rs
@@ -191,6 +191,9 @@ enum ErrorKind {
         available: &'static [&'static str],
     },
 
+    /// Unquoted string was found when quoted one was expected
+    UnquotedString,
+
     #[doc(hidden)]
     __Nonexhaustive,
 }
@@ -1428,7 +1431,7 @@ impl<'a> Deserializer<'a> {
                 start,
                 end,
             },
-            Some((span, Token::Keylike(key))) => self.number_or_date(span, key)?,
+            Some((span, Token::Keylike(key))) => self.parse_keylike(at, span, key)?,
             Some((span, Token::Plus)) => self.number_leading_plus(span)?,
             Some((Span { start, .. }, Token::LeftBrace)) => {
                 self.inline_table().map(|(Span { end, .. }, table)| Value {
@@ -1451,11 +1454,25 @@ impl<'a> Deserializer<'a> {
                         expected: "a value",
                         found: token.1.describe(),
                     },
-                ))
+                ));
             }
             None => return Err(self.eof()),
         };
         Ok(value)
+    }
+
+    fn parse_keylike(&mut self, at: usize, span: Span, key: &'a str) -> Result<Value<'a>, Error> {
+        let lowercase = key.to_ascii_lowercase();
+        let lowercase = lowercase.as_str();
+        if lowercase == "inf" || lowercase == "nan" {
+            return self.number_or_date(span, key);
+        }
+
+        let first_char = key.chars().next().expect("key should not be empty here");
+        match first_char {
+            '-' | '0'..='9' => self.number_or_date(span, key),
+            _ => Err(self.error(at, ErrorKind::UnquotedString)),
+        }
     }
 
     fn number_or_date(&mut self, span: Span, s: &'a str) -> Result<Value<'a>, Error> {
@@ -2076,7 +2093,7 @@ impl std::convert::From<Error> for std::io::Error {
 
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        match self.inner.kind {
+        match &self.inner.kind {
             ErrorKind::UnexpectedEof => "unexpected eof encountered".fmt(f)?,
             ErrorKind::InvalidCharInString(c) => write!(
                 f,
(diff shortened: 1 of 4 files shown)


src/de.rs
```diff
@@ -1451,13 +1454,27 @@ impl<'a> Deserializer<'a> {
...
             }
             None => return Err(self.eof()),
         };
         Ok(value)
     }
 
+    fn parse_keylike(&mut self, at: usize, span: Span, key: &'a str) -> Result<Value<'a>, Error> {
+        let lowercase = key.to_ascii_lowercase();
+        let lowercase = lowercase.as_str();
+        if matches!(lowercase, "inf") || matches!(lowercase, "nan") {
```
alexcrichton: Can this not use `matches!` and use `==` instead?

ehuss:
> failed to parse as a number or datetime, did you mean to use a quoted string?

This sounds good!  Yea, I don't think the error message should say "expected quoted string" because that might be wrong. The parser doesn't know if this value should be a string or not, since there isn't any type information, so softening the message seems good.

My intent about suggesting checking the starting characters was to differentiate between "this is probably not a date or number" vs "this probably is a date or number, but formatted incorrectly".  I'm not sure if checking the starting characters is a good heuristic for that, but I think it would be good to keep the differentiation.  I'm not sure what you mean by "catch parse errors".  Like, instead of always checking the starting characters, let parsing continue normally, and only apply some heuristic if there is an error to decide which error to emit?

src/de.rs
aleksator: Was probably a leftover from some experiments, it should definitely use ==.

aleksator:
>  I'm not sure what you mean by "catch parse errors". Like, instead of always checking the starting characters, let parsing continue normally, and only apply some heuristic if there is an error to decide which error to emit?

This is a good question.
I will reword the message, but will keep the implementation until there is a confirmation that it is indeed a desirable behavior.

aleksator:
Actually after reading through Alex's answer a couple more times it seems that the behavior should be changed to checking the error and not using the heuristic. I'll get to work on that.

src/de.rs
```diff
@@ -1451,13 +1454,32 @@ impl<'a> Deserializer<'a> {
...
-                ))
+                ));
             }
             None => return Err(self.eof()),
         };
         Ok(value)
     }
 
+    fn parse_keylike(&mut self, at: usize, span: Span, key: &'a str) -> Result<Value<'a>, Error> {
+        self.number_or_date(span, key).map_err(|_| {
```
alexcrichton: Oh sorry I don't mean to say that *all* errors should be translated to unquoted strings here, because some could legitimately be malformed numbers or something like that. I think it's probably just helpful to append/augment the existing error to indicate "you may have meant to use a string here"

aleksator: Is the new implementation closer to what you had in mind?

aleksator:
The latest message is this:
`"invalid number, did you mean to use a quoted string \"9223372036854775808\"? at line 1 column 5"`
Maybe it'd better to shift the order a bit:
`"invalid number \"9223372036854775808\", did you mean to use a quoted string? at line 1 column 5"`

What do you guys think?

alexcrichton:
I think the error message is improving, but sorry I think this is still a bit too overzealous about warning about whether a quoted string was meant or not. I think it might be best to do something like see if the parse failed on the first character, since that's what's likely to happen for `foo = bar` or something like that. Additionally I don't think that the error handling code should have `.expect()` in it.

aleksator:
Getting back to finishing the pull requests that have been hanging for a while.

I have returned the heuristic back and changed the error message to match the suggested one exactly.

@alexcrichton Please take a look if you'd like anything else changed whenever you have some free time. The difference between this and master is quite small in this version.

aleksator:
@alexcrichton @ehuss Could you please take a look if any other changes are needed here?

By the way, one of the CI jobs never started, and I don't know how to restart it myself without doing superficial changes.

src/de.rs
```diff
@@ -1451,13 +1454,27 @@ impl<'a> Deserializer<'a> {
...
             }
             None => return Err(self.eof()),
         };
         Ok(value)
     }
 
+    fn parse_keylike(&mut self, at: usize, span: Span, key: &'a str) -> Result<Value<'a>, Error> {
+        let lowercase = key.to_ascii_lowercase();
+        let lowercase = lowercase.as_str();
+        if lowercase == "inf" || lowercase == "nan" {
```
ehuss, line 1467: Just a quick question:  What is the reason to use lowercase here?  It's probably not too expensive, but it seems like an unnecessary expense for an obscure edge case.

ehuss (commented):
Sorry for the long delay!

src/de.rs
aleksator: I though all of the spellings are valid and wanted to check those.

Checked the spec:
> Special float values can also be expressed. They are always lowercase.

https://toml.io/en/v1.0.0-rc.3#user-content-float
I was wrong and I must not convert strings to lowercase. Will fix.

aleksator: Fixed.

ehuss (approved):
Thanks!  I'm still not sure about the exact wording, since I imagine it could still be potentially confusing for someone who doesn't understand TOML. But it is a definite improvement!

I'll leave this open for a bit if @alexcrichton wants to chime in.

aleksator:
> Thanks! I'm still not sure about the exact wording, since I imagine it could still be potentially confusing for someone who doesn't understand TOML. But it is a definite improvement!

I'm up to change it if you have some suggestions!

ehuss:
> I'm up to change it if you have some suggestions!

Unfortunately I don't think I have any better ideas.  It is tough to craft an error message that isn't confusing.  My only thought is that if the user writes `foo = bar`, and the error message talks about numbers and datetimes, the user may be confused about why since they are clearly not writing a number.

Maybe something like `invalid TOML value, did you mean to use a quoted string?`
I dunno if that is more or less of clear.

aleksator:
> Maybe something like invalid TOML value, did you mean to use a quoted string?
I dunno if that is more or less of clear.

Agree, that does sound clearer and doesn't leak parsing order. Changed.

alexcrichton:
Sorry for being so slow to move on this, but this looks good to me!

Issue #389, opened 2020-04-30
#![no_std] support


diondokter:
I would like to use toml for some embedded purposes.
Looking through the code, I don't see anything that would prevent the support.

Both IndexMap and Serde support it.

PR #391, opened 2020-05-09
Added no_std support

diondokter:
Following up on #389 

This PR adds the ability to use the library in a no_std context.
Tested it on my embedded setup and it works as expected.

Consequences:
- Extra feature added as default: `std`
- Imports are more complicated.
- Instead of using the std hashmap, the hashbrown hashmap is used. This shouldn't have any effects because it is the same hashmap that is used in the std.
Ideally https://github.com/rust-lang/rust/issues/56192 would be made.

The imports could be cleaned up. The Serde crate does this in its lib.rs, but I don't think that's really necessary.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,6 +12,7 @@ jobs:
     - uses: actions/checkout@master
     - name: Install Rust (rustup)
       run: rustup update ${{ matrix.rust }} && rustup default ${{ matrix.rust }}
+    - run: cargo build --no-default-features
     - run: cargo test
     - run: cargo test --features preserve_order
     - run: cargo test --manifest-path test-suite/Cargo.toml
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,15 +20,17 @@ edition = "2018"
 members = ['test-suite']
 
 [dependencies]
-serde = "1.0.97"
+serde = { version = "1.0.97", default-features = false, features = ["alloc"] }
 indexmap = { version = "1.0", optional = true }
+hashbrown = { version = "0.7.2" }
 
 [dev-dependencies]
 serde_derive = "1.0"
-serde_json = "1.0"
+serde_json = { version = "1.0", default-features = false, features = ["alloc"] }
 
 [features]
-default = []
+default = ["std"]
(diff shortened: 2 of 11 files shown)


alexcrichton:
Thanks for the PR, but I'm personally not too thrilled to add `no_std` support like this. To me `no_std` means "only `core`", not "both `core` and `alloc`". This library gets pretty un-idiomatic pretty quickly (as shown here) if `alloc` enters the picture, and I do not want to maintain the `alloc`-specific code.

diondokter:
@alexcrichton 

I'm using it for reading a config file from an SD card in an embedded system.
I don't think you'll want to parse toml on low resource systems where you can't have a decent heap as the deserialize function for my config type takes up a lot of flash.

image

While it's true that it can definitely be better, this makes it usable instead of unusable.
Not using alloc anywhere, though, would be a really big change.
- No Strings
- No Hashmaps
- No Vecs
- No fmt

But I do get that you don't want to maintain it as I don't expect a great many people are going to use it.

So if you're not going to to add support in this form or another, would you mind me creating a fork that has it and put it up on crates.io? I don't know how well I could maintain it, but at least there'd be the option, just like serde_json_core is/was and you don't have to maintain it. Of course I'd write in the description that you should probably use this normal crate.

alexcrichton:
Yes not using `alloc` is I believe practically impossible with this library as-is. If you'd like to fork this for this use case that's ok by me.

YtvwlD:
I'm currently thinking about writing a UEFI application and this would have been really useful.
@diondokter are you planning to put this up on crates.io?

diondokter:
@YtvwlD no, not really as I move to something else (non-toml). But my fork is still up. Feel free to fork it yourself and put it up on crates.io 🙂

YtvwlD:
I'm not *that* keen on maintaining a fork, but let's see. I already tried your fork and it worked.

What did you switch to (if I'm allowed to ask)? I'm also considering JSON but I'd like to have comments in a configuration file.

diondokter:
I moved to a sort of registry format because that suited my needs better. So a key/value storage.

clearloop:
What about an `alloc` feature?

Issue #387, opened 2020-04-14
Better handling for quotes in multiline strings


ehuss:
The spec clarified some issues about how quotes are handled in multiline strings, and I believe the lexer needs to be updated, as some of the examples fail.

More information at:
* https://github.com/toml-lang/toml/pull/648 — Changes to ABNF
* https://github.com/toml-lang/toml/issues/725 — Examples of various edge cases
* https://github.com/toml-lang/toml/pull/727 — Updates with some examples

ehuss:
Closed by #393.

Issue #392, opened 2020-05-14
Update to v1.0.0-rc.1


pyfisch:
On April 1 a new version of TOML was published: https://github.com/toml-lang/toml/blob/master/CHANGELOG.md#100-rc1--2020-04-01

The changelog mentions the following changes:

* [x] Clarify in ABNF how quotes in multi-line basic and multi-line literal strings are allowed to be used. (PR in #393)
* [x] Leading zeroes in exponent parts of floats are permitted.
* [x] Clarify that control characters are not permitted in comments.
* [ ] Clarify behavior of tables defined implicitly by dotted keys.
  * [x] spec text:
    > Dotted keys define everything to the left of each dot as a table. Since tables cannot be defined more than once, redefining such tables using a `[table]` header is not allowed. Likewise, using dotted keys to redefine tables already defined in `[table]` form is not allowed.
  * [ ] spec text ([example file][4]):
    > The `[table]` form can, however, be used to define sub-tables within tables defined via dotted keys.
* [x] Clarify that inline tables are immutable.
* [x] Clarify that trailing commas are not allowed in inline tables.
* [x] Clarify in ABNF that UTF-16 surrogate code points (U+D800 - U+DFFF) are not allowed in strings or comments. (*ensured by Rust*)
 * [x] Allow raw tab characters in basic strings and multi-line basic strings.
 * [x] Allow heterogenous values in arrays. (was changed in #358)

TOML is getting [official compliance tests][1]. The repository contains no actual tests yet but a script to run the compliance tests eventually. I think the best way to reach v1.0 and to avoid bugs is to use the test suite to check both deserialization and serialization.

While this crate already contains a test runner for an almost identical format it may be good to interface with the default Python test runner. If many implementations use the default runner it becomes easy to monitor the correctness of many different implementations.

* [ ] Write script implementing the [decoder interface][2] (*a serializer for the JSON format already exists*) <https://github.com/alexcrichton/toml-rs/blob/7a43ac9e9c28afec8414f6d924024f5d33fe8fa2/test-suite/tests/valid.rs#L9-L54>
* [ ] Write script implementing the [encoder interface][3]
* [ ] **Optional**: Remove rust test runner (upside: less code, downside: Python is needed to test this crate)

[...]

PR #393, opened 2020-05-24
Allow delimiter quotes at the end of multiline strings

pyfisch:
TOML allows (unlike many other formats) up to 2
additonal quotes that are part of the string:

```
basic = """2 extra quotes -->"""""
literal = '''here too ''''
```

Changed in TOML v1.0.0-rc.1

See also #392

### DIFF
--- a/src/tokens.rs
+++ b/src/tokens.rs
@@ -334,7 +334,7 @@ impl<'a> Tokenizer<'a> {
                         return Err(Error::NewlineInString(i));
                     }
                 }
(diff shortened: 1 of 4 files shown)


alexcrichton:
I believe this is handling https://github.com/toml-lang/toml/pull/648, right? If so looks great to me, thanks!

pyfisch:
Yes this is toml-lang/toml#648. Also it resolves #387.

alexcrichton:
👍

PR #396, opened 2020-06-11
Introduce new toml_lexer crate

### LINKED ISSUE #394: Extract a self-contained toml_lexer crate
Hi, I'd like to use your `lexer` in the implementation of the [language server for Cargo.toml](https://github.com/matklad/tom).
This is the thing we have already done with `rustc_lexer` (i.e. `rustc_lexer` is shared between `rustc` and `rust-analyzer`), so I'd like to do a similar thing here too.

@alexcrichton I know you are busy, but I am willing to help. If I restructure your repo to a cargo workspace with `crates/toml_lexer` and `crates/toml`, such that you can publish the lexer, would you accept this change? If you don't have time for publishing or even maintaining this crate, please consider giving the maintenance rights to any contributor you trust here, or me :D

Veetaha:
The documentation on the crate speaks for itself.
Shortly this crate's main idea is to be very generic to support multiple
use cases for toml tokenizer (i.e. `toml-rs` serde parsing itself and
error-resilient parsing for some programs like TOML language servers).

The crate exposes very low-level api, even further sub-tokenizing string
literals. The api and implementation is highly inspired by `rustc_lexer`,
and its ideas fully correspond.

I have a self-interest in maintaining `toml_lexer` crate, I assume I will need utilities for parsing toml numbers and dates from `Keylike` token somewhere in near future, but I didn't bother implementing it in this PR just to keep the number of changes much lower (which is to say quite large, sorry for so many changes, even tho @alexcrichton supposed it would be possible to implement error-resilient tokenizer on top of the already existing one with small changes, but I have what I have, I iterated on the api and implementation quite a lot and I couldn't introduce fewer changes...)

I also organized all crates under the `crates/*` dir, since `toml` and `toml_lexer` are now fully independent crates, if you find this wrong, I'll return it to the previous layout and add `toml_lexer` as a top-level dir.

Closes #394


alexcrichton:
Awesome and thanks for this! I started taking a look at this and it's looking pretty good to me in terms of architecture. One thing I wanted to ask about though was how low-level this is. The tokenizer for serde, for example, seems to have only reduced in size slightly in that for the serde purpose consume the raw output of the lexer crate still requires quite a bit of processing. Is this sort of processing going to need to be duplicated amongst users of this crate? 

One of the main goals of splitting this out was to share logic across various consumers of TOML, but I'd be slightly afraid if the real meat of the logic is still within the toml crate itself. It looks like the main part separated is the string lexing, but other than that the new crate doesn't seem to do a huge amount?

Veetaha:
Well, the tokenizer for `serde` has quite a lot of additional private helper methods which I haven't moved into `toml_lexer` since these are very ad-hoc for serde sphere.
Other than that mapping between `toml_lexer` possibly-malformed-tokens and original `toml` crate tokens takes some lines of code, so it may seem like there is a still lot of logic in `tokens.rs`.

And you are right, the gist of logic is dedicated to string literals sub-tokenizing (which is very useful e.g. for highlighting escape sequences inside of string literals for language server use case). I had a thought to move making `StrLitSubtoken`s into a `Cow<str>` to `toml_lexer` crate but this requires moving the error enum to `toml_lexer` crate, and this, in turn, means we would need to maintain that as a public API.

Well, one of the goals I had was making the public API surface as small as possible, but if you don't mind moving collecting string literals into `Cow<str>` to `toml_lexer` it's hard to peek a robust api for this one.

Anyway, I do have one idea (I did a trial impl to check, and it should work) 
This will be more verbose but will fully leverage the type system and ownership model by encoding the state into the types further:

```rust
// toml_lexer
struct Token<K> { span: Span, kind: K }

enum TopLevelTokenKind {
    Comment, 
    Keylike, 
    // ... all trivial top-level tokens (non-string-literals)
}

enum StrLitSubtokenKind {
    Char(char),
    UnicodeEscape(UnicodeEscape),
    // ... all str lit subtokens exlcuding leading and trailing quotes
}

struct TopLevelTokenizer(...);
impl TopLevelTokenizer {
    // notice consuming self
    fn next(self) -> Option<TopLevelNextState>;
}
enum TopLevelNextState {
    TopLevel(Token<TopLevelTokenKind>, TopLevelTokenizer),
    LeadingQuotes(Token<Quotes>, StrLitTokenizer),
}

struct StrLitTokenizer(...);
impl StrLitTokenizer {
    fn next(self) -> StrLitNextState;
    /// Consumes self and returns the next state with the fully unescaped string if it is well-formed.
    fn unescape_into_cow(self) -> Result<(Cow<str>, TopLevelTokenizer), EscapeError>;
}
enum StrLitNextState {
    // Optional trailing quotes span
    TopLevel(Option<Span>, TopLevelTokenizer),
    StrLit(Token<StrLitSubtokenKind>, StrLitTokenizer),
}


// tom/src/tokens.rs

struct SerdeTokenizer(Option<toml_lexer::TopLevelTokenizer>);
impl SerdeTokenizer {
    fn new(input: &str) -> Self {
        Self(Some(TopLevelTokenizer::new(input)));
    } 

    fn next() -> Result<Option<(Span, Token)>, Error> {
       let tokenizer =  self.0.take()?;
       match tokenizer.next()? {
           TopLevelNextState::LeadingQuotes(quotes, str_lit_tokenizer) => {
                let (unescaped, top_level_tokenizer) = str_lit_tokenizer
                    .unescape_into_cow()
                    .map_err(|err| Error::EscapeError(err))?;

                self.0 = Some(top_level_tokenizer);
                String {
                    val: unescaped,
                    multiline: quotes.kind == Quotes::X3,
                }                
           }
           // ... other  token mappings are trivial 1-to-1
       }
}
```

I love this because this decreases the number of `unwrap()`, but I hate this because it moves us from concise `impl Iterator` to verbose `fn next(self) -> NextStateEnum`.

Veetaha:
I think I'll stick with simple iterator approach, but add a method
```rust
fn eat_str_lit(&mut self) -> Option<Result<Cow<str>, StrLitError>>
```
I also have minor changes to `tokenize_str_lit()` api in mind (returning `usize` instead of `Span` there just because the returned span always starts from `0`).
Besides that I am already working on parsing number and date literals, I won't include this work in this particular PR (rather in a separate followup PR). I think it will look similarly:

```rust
enum NumOrDatetimeLitResult {
    Num(Result<Number, NumError>),
    Datetime(Result<Datetime, DatetimeError>),
}

fn eat_num_or_datetime_lit(&mut self) -> Option<NumOrDatetimeLitResult>
```


I actually think I will test my work in `tom` first to have a real test case for `toml_lexer` api and then move this PR from draft.

Veetaha:
Unfortunately, I am bit out of capacity to make this. I'll close the PR, but anyone else interested is free to continue the work!

PR #416, opened 2021-01-23
Enforce millisecond precision for DateTime

aduh95:
TOML 1.0 spec requires millisecond precision for:

- Offset Date-Time
- Local Date-Time
- Local Time

Refs: https://toml.io/en/v1.0.0#offset-date-time

> Millisecond precision is required. Further precision of fractional seconds is implementation-specific. If the value contains greater precision than the implementation can support, the additional precision must be truncated, not rounded.


dtolnay (changes requested):
I think this change is wrong and the previous behavior is correct. This PR seems to make `00:00:00.00` invalid syntax because it doesn't have at least 3 digits in the fractional seconds part. The toml grammar is quite clear that only at least 1 digit is required: https://github.com/toml-lang/toml/blob/8eae5e1c005bc5836098505f85a7aa06568999dd/toml.abnf#L173. See https://www.ietf.org/rfc/rfc5234.txt for the meaning of `"." 1*DIGIT`.

The prose that you quoted is a constraint on the implementation, not on the input format. A conformant toml implementation is not allowed to treat `00:00:00.00` as identical to `00:00:00.01`, but is allowed to treat `00:00:00.0000` as identical to `00:00:00.0001` because submillisecond precision is optional and implementation-specified.

aduh95:
OK makes sense, the ABNF document is indeed very clear. I'll open a PR on the TOML spec to see if we can clarify the wording.

PR #422, opened 2021-04-02
Use `constant` instead of `static` in a couple places

kulakowski:
This keeps symbols for these from accidently being exported as easily.

### DIFF
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -345,7 +345,7 @@ impl<'de> de::Deserialize<'de> for Datetime {
             }
         }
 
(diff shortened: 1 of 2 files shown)


kulakowski:
Well, a `-0` vs `0` serialization difference on nightly that doesn't reproduce for me locally is something I don't want to think about too hard to start a weekend. I will hopefully dig into this soon.

xpe:
"This keeps symbols for these from accidently being exported as easily." -- kulakowski

Is this PR trying to avoid an anti-pattern? (I don't think I'm familiar with it.) @kulakowski Could you explain a little more?

epage:

PR #424, opened 2021-04-03
Introduce deserialize_datetime_as_string feature flag

### LINKED ISSUE #423: Add deserialize-datetime-as-string feature flag
The `Deserializer` of this crate deserializes TOML datetimes as a map with a magic string, so e.g. `2021-04-03` becomes `{"$__toml_private_datetime" = "2021-04-03"}`. While the serializer of this crate catches these magic strings and converts them back into TOML datetimes (which is apparently necessary because serde provides no way to cleanly communicate types outside of its datamodel, see serde-rs/serde#1643), serializers from other crates obviously do not know about this magic string, resulting in it becoming part of their output if they are used with this crate.

I therefore propose the introduction of an optional `deserialize-datetime-as-string` feature flag (which would be disabled by default). If enabled this flag would make the `Deserializer` yield datetimes as strings instead of maps, making it for example easy to convert TOML to JSON without risking the leaking of this magic string.

If you think this is a good idea, I'd be happy to work on a PR.

Relevant comments in the code:

https://github.com/alexcrichton/toml-rs/blob/b1417006df376e6c406bff43740dd11aa7ef744e/src/datetime.rs#L37-L44

not-my-profile:
When using the Deserializer implementation of this crate
with a serde::de::Deserialize implementation of another crate,
you don't want e.g. 2021-04-03 to become

{"$__toml_private_datetime" = "2021-04-03"}

(which is the special encoding this crate uses to represent
datetimes since serde's data model lacks a type for datetimes)

This commit introduces an off by default feature flag to deserialize
datetimes as strings if so desired, making it feasible to use
the toml crate with other Deserialize implementations without
leaking the above internal representation.

Closes #423.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,3 +34,9 @@ default = []
 # This allows data to be read into a Value and written back to a TOML string
 # while preserving the order of map keys in the input.
 preserve_order = ["indexmap"]
+
+# If you use toml::de::Deserializer with a serde::de::Deserialize
+# implementation of another crate, you can enable the following flag to
+# deserialize datetime values as strings, instead of the special encoding
+# this crate normally uses behind the scenes.
+deserialize_datetime_as_string = []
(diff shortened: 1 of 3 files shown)


not-my-profile:
Closing this since it's not a good solution since it changes the interface of the type.

PR #429, opened 2021-05-12
Support use without libstd (e.g. alloc + core only)

thomcc:
This allows using toml for config files in some scenarios where e.g. porting libstd isn't really viable or would be painful, but where a small memory allocator is easy.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,6 +14,8 @@ jobs:
       run: rustup update ${{ matrix.rust }} && rustup default ${{ matrix.rust }}
     - run: cargo test
     - run: cargo test --features preserve_order
+    - run: cargo test --no-default-features
+    - run: cargo test --no-default-features --features preserve_order
     - run: cargo test --manifest-path test-suite/Cargo.toml
     - run: cargo bench
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ edition = "2018"
 members = ['test-suite']
 
 [dependencies]
-serde = "1.0.97"
+serde = { version = "1.0.97", default-features = false }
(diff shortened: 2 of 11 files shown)


alexcrichton:
Thanks for the PR, but I personally do not want to maintain such support. If you have targets you're interested in building which are missing `std` then I'd recommend getting `std` to build for those targets instead.

thomcc:
> If you have targets you're interested in building which are missing std then I'd recommend getting std to build for those targets instead

Of course, doing this is not remotely always viable. A memory allocator can be implemented in a few hundred lines of code — implementing support for the operations libstd needs requires far more than this. That's why core+alloc is separate in the first place.

> I personally do not want to maintain such support

I mean, of course that's your call as maintainer, but given that doing so amounts to very little more than shuffling the imports around some, it feels like an ideological stance, which is, well, frustrating, to be honest. :/

alexcrichton:
Supporting libstd doesn't mean implementing and supporting everything like threads, it just means getting it to compile for your target. It's always possible for libstd to simply not export, for example, a `threads` module on platforms that don't support threads. Alternatively it's fine for functions to be exposed, they just always return errors on platforms where threads aren't implemented yet.

I think both of those alternatives are better than trying to change every crate in the ecosystem to not use `std`.

thomcc:
I'm aware, but that's still way more work than getting a few packages which don't support no_std usage but easily can to use liballoc instead.

It also moves Rust towards the undesirable state C/C++ is in where on embedded some unknown portion of the stdlib works, and possibly the portion that works is buggy... The current situation isn't ideal, but IMO it's a better starting point than what we have for C/C++.

Not to mention, it's likely far easier to convince projects to use no_std+alloc than it is to avoid some specific libstd modules that you don't happen to have access to on your embedded system of choice.

PR #435, opened 2021-07-29
Fix missing empty line between tables

### LINKED ISSUE #219: Formatting of map after struct
There is no empty line inserted when a map is serialized after a struct. This behaviour is inconsistent to the formatting of other table-like structures, i.e. structs and arrays. It's not a big deal as it does not violate the TOML specification, but it may impede readability.

### Minimal Example:
```rust
extern crate serde;
#[macro_use] extern crate serde_derive;
extern crate toml;

use std::collections::HashMap;


#[derive(Serialize, Deserialize)]
struct Foo {
    bar: Bar,
    baz: HashMap<String, Bar>,
}

#[derive(Serialize, Deserialize)]
struct Bar {
    foo: u32
}

fn main() {
    let mut baz = HashMap::new();
    baz.insert("a".into(), Bar { foo: 0 });

    let foo = Foo {
        bar: Bar {
            foo: 42,
        },
        baz: baz,
    };

    println!("{}", toml::to_string_pretty(&foo).unwrap());
}
```

#### Actual Output:
```toml
[bar]
foo = 42
[baz.a]
foo = 0
```

#### Expected Output:
```toml
[bar]
foo = 42

[baz.a]
foo = 0
```

Swapping the `bar` and `baz` members of `Foo` produces an expected result.

Kixunil:
This adds empty line that was missing between tables, specifically
between struct and map. It also adds test to catch those cases.

Closes #219

### DIFF
--- a/src/ser.rs
+++ b/src/ser.rs
@@ -934,10 +934,15 @@ impl<'a, 'b> ser::Serializer for &'b mut Serializer<'a> {
 
     fn serialize_map(self, _len: Option<usize>) -> Result<Self::SerializeMap, Self::Error> {
         self.array_type(ArrayState::StartedAsATable)?;
+        let first = match self.state {
+            State::Table { first, .. } => first.get(),
+            State::Array { first, .. } => first.get(),
+            State::End => true,
+        };
         Ok(SerializeTable::Table {
             ser: self,
             key: String::new(),
-            first: Cell::new(true),
+            first: Cell::new(first),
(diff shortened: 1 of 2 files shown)


ehuss:
There appears to be an error in CI.  Do you want to look at this further?

Kixunil:
Ah, I missed it, thanks for heds up! I will.

epage:

PR #438, opened 2021-08-09
Use entry api to remove a lookup in deserialisation loop

gilescope:
Was reading the source and this jumped out at me.
It's not much but deserialising toml faster seems like a good thing.


alexcrichton:
Thanks! I think a rustfmt is needed though?

gilescope:
Good point. Sorted.

PR #441, opened 2021-08-21
Partial ordering for `Datetime`

xpe:
First, would the maintainers be interested in accepting this kind of addition?

## Summary

This is a work-in-progress. So far, I've implemented `PartialOrd` for `Date` and `Time`. I've started with `Datetime`, and I wanted to outline my plan to get feedback.

*The Plan*: The only possibility for a non-`None` ordering happens between TOML *Offset Date-Time*s (i.e. cases where the `Datetime` struct has no `None` values). To compare such as pair (`self` and `other` in the `partial_cmp` function), I plan to convert each to its UTC equivalent. Is this correct in all cases?

### DIFF
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -61,6 +61,7 @@ pub struct Time {
     pub nanosecond: u32,
 }
 
+#[allow(missing_docs)]
 #[derive(PartialEq, Clone)]
 pub enum Offset {
     Z,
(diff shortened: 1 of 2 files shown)


xpe:
I'll reopen this PR when the branches are cleaned up.

src/datetime.rs
```diff
@@ -92,6 +94,72 @@ pub struct Datetime {
...
+        (self.date.is_none() && self.offset.is_some()) ||
+        (self.time.is_none() && self.offset.is_some())
+    }
+}
+
+impl PartialOrd for Datetime {
+    /// Across the 10 pairwise combinations of valid `Datetime` types,
+    /// there is only one combination that has an ordering defined.
+    /// It is a pair of two TOML Offset Date-Times.
+    A partial ordering only exists between TOML Offset Date-Times.
```
xpe, line 153: Oops. This snuck in.

PR #442, opened 2021-08-21
Implement Debug for Date, Time, Offset

xpe:
I see that many other structs implement Debug in this way. How does this look?

(This will help a downstream project I'm working on.)

### DIFF
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -207,9 +207,21 @@ impl fmt::Display for Datetime {
     }
 }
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
I think these might be a bit better derived rather than forwarding to display? That way it should be clear which field is which I think

xpe:
I'll take a look.

not-my-profile:
I made the structs derive Debug in #455.

epage:

PR #443, opened 2021-08-21
Partial ord datetime

xpe:
Would the maintainers be interested in accepting this kind of PR...?

## Summary

I've implemented `PartialOrd` for `Date` and `Time`. I've started but not finished doing the same for `Datetime`. I'd like to outline my plan to get feedback.

*The Plan*: The only possibility for a non-`None` ordering happens between TOML *Offset Date-Time*s (i.e. cases where the `Datetime` struct has no `None` values). To compare such a pair (in `partial_cmp`), I plan to convert each to its UTC equivalent first. From there, a comparison is easy. Is this correct in all cases?

P.S. Apologies for an earlier duplicate; I closed it because it was based on the wrong branch.

### DIFF
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -207,9 +207,21 @@ impl fmt::Display for Datetime {
     }
 }
 
+impl fmt::Debug for Date {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt::Display::fmt(self, f)
+    }
+}
+
 impl fmt::Display for Date {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "{:04}-{:02}-{:02}", self.year, self.month, self.day)
+    }
+}
+
+impl fmt::Debug for Time {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt::Display::fmt(self, f)
     }
 }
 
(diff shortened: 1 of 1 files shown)


xpe:
> From there, a comparison is easy.

Famous last words.

xpe:
Here in toml-rs, the `Datetime` struct covers four kinds of TOML datetimes:

* ODT = Offset Date-Time
* LDT = Local Date-Time
* LD = Local Date
* LT = Local Time

It may be useful to know that there are 10 combinations of how a `Datetime` might be compared. A table shows this most clearly (because order doesn't matter in a combination):

|        | ODT    | LDT    | LD     | LDT    |
| ------ | ------ | ------ | ------ | ------ |
| ODT    | 1      | 2      | 3      | 4      |
| LDT    | 2      | 5      | 6      | 7      |
| LD     | 3      | 6      | 8      | 9      |
| LT     | 4      | 7      | 9      | 10     |

## Partial Ordering?

Support for this interpretation comes from:

* The TOML 1.0.0 specification about [Offset Date-Time]s:

> "To unambiguously represent a specific instant in time, you may use an RFC 3339 formatted date-time with offset."

* ... [Local Date-Time]

> Conversion to an instant, if required, is implementation-specific.

## My (current) interpretation

Here, in the context of the toml-rs library, I interpret the above to mean that **only** `Datetime`s that have an offset are comparable. As such, only <span>#</span>1 in the table above (ODT vs. ODT) above may be ordered.

Conceptually, without knowing what "local" means for a particular TOML file, it is impossible to compare a local `Datetime` with an offset `Datetime`. Applications would need to handle these cases.

[Offset Date-Time]: https://toml.io/en/v1.0.0#offset-date-time
[Local Date-Time]: https://toml.io/en/v1.0.0#local-date-time

alexcrichton:
Personally I would prefer to not be in the business of having lots of support for a datetime type. I'd rather than the type was easily convertible to something like `chrono` and various comparison operators from there could be used instead

xpe:
@alexcrichton Makes sense. I'll try to circle back and check on converting to `chrono`.

PR #445, opened 2021-09-23
Added the ability to deserialise signed and unsigned 128 bit integers.

TheDarkula:
I have verified the functionality with this sample code:
```
use serde::{Serialize, Deserialize};

#[derive(Deserialize)]
struct Config {
    signed: Option<i128>,
    unsigned: Option<u128>,
}

#[derive(Deserialize)]
struct Keys {
    github: String,
    travis: Option<String>,
}

fn main() {
    let config: Config = toml::from_str(r#"
        signed = 333
        unsigned = 333
    "#).unwrap();

    println!("{}", config.signed.unwrap());
    println!("{}", config.unsigned.unwrap());
}
```

Also, you will need this in your `Cargo.toml`:
```
[dependencies.serde]
version = "1.0.130"
features = ["derive"]

[dependencies]
toml = { path = "../../toml-rs-thedarkula" }
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "toml"
-version = "0.5.8"
+version = "0.6.0"
 authors = ["Alex Crichton <alex@alexcrichton.com>"]
 license = "MIT/Apache-2.0"
 readme = "README.md"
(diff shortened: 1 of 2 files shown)


TheDarkula:
It has come to my attention that this only functions properly if the number is small.

Using this:
```
    let config: Config = toml::from_str(r#"
        signed = 18446744073709551616
        unsigned = 18446744073709551616
    "#).unwrap();
```
causes a failure.

I am looking further into a proper implementation.

TheDarkula:
Unfortunately, given that all integers are read as an `i64`, adding this functionality will take quite a bit of a rewrite.

PR #447, opened 2021-10-21
Improve dotted table parsing

HarveyHunt:
## Context
#225 was filed to request externally tagged enum support, which was mostly implemented in #264. However, #264 didn't have complete support for enums in dotted tables. The PR did add some disabled tests to verify the functionality in the future.

## Examples
As shown in the test file, before this change the following TOML would fail to parse:

```
[val]
NewType = "value"
```

Here's an excerpt of the error message: ```"unknown variant `val`, expected one of `Plain`, `Tuple`, `NewType`, `Struct`", key: ["val"]```

The code was also unable to parse an array of tables into enum values, such as:

```
[[enums]]
Plain = {}

[[enums]]
Tuple = { 0 = -123, 1 = true }

[[enums]]
NewType = "value"

[[enums]]
Struct = { value = -123 }
```

Here's an excerpt of the error message: ```"enum table must contain exactly one table", key: ["enums"] ```

Implement the missing functionality and enable the test cases that were previously added.

---

Note: This PR is functionally identical to #362, but is rebased and has an extra commit that removes the `FIXME`s  and a lint. @johansglock and I have chatted offline about this and agreed that I'll submit this new PR in order to get this change upstream, as he has limited time at the moment.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -11,6 +11,7 @@ use std::f64;
 use std::fmt;
 use std::iter;
 use std::marker::PhantomData;
+use std::ops::Deref;
 use std::str;
 use std::vec;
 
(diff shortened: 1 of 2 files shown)


HarveyHunt:
@ehuss @alexcrichton, would you be able to take a look at this PR please? This feature has been requested quite a few times in the original PR (#362). Thanks!

ehuss:
@HarveyHunt Would you be willing to post a separate PR to fix the CI issues?  I think it would just require sprinkling some `#[allow(dead_code)]` where necessary.  That is a prerequisite for getting things rolling again.

HarveyHunt:
@ehuss, I can send a PR for that!

ehuss:
So just rethinking about this at a high level, there are a few surprising behaviors on how enums are handled. I'm just jotting some notes down here, this isn't a review. I'll try to go through the code itself later. IIUC, this isn't intended to fully support enum variants completely, just to try to push a few more cases through. It also seems like serialization is not supported at all, which is a little surprising. (These notes don't need to be addressed, I'm just trying to capture some things I've noticed.)

## Fieldless variants

Fieldless variants seem to require an empty table, like:

```toml
[val]
Plain = {}
```

My initial thought is that it should mirror whatever serde_json does, as in:

```toml
val = "Plain"
```

This PR now allows the former.  I guess I don't feel strongly about it, but it seems strange to not follow the same behavior as other serde deserializers.

## Tuple variants

Tuple variants don't seem to work as I expect:

```toml
[val]
Tuple = [1, true]
```

It does allow it with numeric fields:

```toml
[val]
Tuple = {0 = 1, 1 = true}
```

I take it this was the style chosen in #264 because at the time TOML didn't allow mixed types in arrays, but it does now.

## Dotted tables

Various cases don't support dotted tables where I would expect:

```toml
val.Struct.value = 123
val.ExternalStruct.key1 = "a"
val.ExternalStruct.key2 = "b"
```

src/de.rs
```diff
@@ -667,34 +665,52 @@ impl<'de, 'b> de::Deserializer<'de> for MapVisitor<'de, 'b> {
...
         let name = table.header[table.header.len() - 1].1.to_owned();
-        visitor.visit_enum(DottedTableDeserializer {
-            name,
-            value: Value {
-                e: E::DottedTable(values),
-                start: 0,
-                end: 0,
-            },
-        })
+        let mut values: Vec<TablePair<'_>> = if self.values.peek().is_some() {
```
ehuss, line 678: Can this have a comment explaining why it is needed? It is not clear to me what scenarios these two different cases would encounter.

src/de.rs
```diff
@@ -667,34 +665,52 @@ impl<'de, 'b> de::Deserializer<'de> for MapVisitor<'de, 'b> {
...
+        if variants.contains(&name.deref()) {
+            visitor.visit_enum(DottedTableDeserializer {
+                name,
+                value: Value {
+                    e: E::DottedTable(values),
+                    start: 0,
+                    end: 0,
+                },
+            })
+        } else if values.len() == 1 {
```
ehuss, line 693: Can this include a comment for the cases `len=1` is covering?  From what I understand, it is trying to handle the newtype situation where `SomeVariant(String)` is being handled as `SomeVariant = "value"`?

HarveyHunt:
Hey @ehuss, sorry it's taken me so long to address your feedback! Life got busy and then I completely forgot about this PR until I spotted it again a couple of days ago.

I've added the comments that you mentioned, let me know if there's anything else that I should add. :+1:

HarveyHunt:
Hey @alexcrichton , @ehuss - would you be able to take a look at this PR please? Thanks!

alexcrichton:
Sorry but I don't have the time/energy nowadays to review feature additions to this crate.

epage:

PR #450, opened 2021-11-08
Allow dead code in examples to fix CI build

HarveyHunt:
The examples deny warnings, however a couple of examples deserialise TOML into
structs. The fields of these structs are never read, which breaks cargo test
in nightly mode.

Allow dead code to fix the CI build, as discussed on #447.


alexcrichton:
Thanks!

PR #452, opened 2021-12-06
 Add `from_reader` and `to_writer` methods

WhyNotHugo:
These match what other serde serialisers offer.

This PR is mostly a rebase of https://github.com/alexcrichton/toml-rs/pull/257, with just one extra commit.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -6,9 +6,11 @@
 
 use std::borrow::Cow;
 use std::collections::HashMap;
+use std::convert::From;
 use std::error;
 use std::f64;
 use std::fmt;
+use std::io;
 use std::iter;
 use std::marker::PhantomData;
 use std::str;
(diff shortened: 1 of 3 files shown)


alexcrichton:
Could these methods perhaps return a new error enum instead of fitting an `Arc` and `io::Error` into the existing one?

WhyNotHugo:
Done. That was actually much simpler than I'd anticipated.

src/de.rs
```diff
@@ -82,13 +84,57 @@ where
...
+    T: de::DeserializeOwned,
+{
+    let mut buf = Vec::new();
+    r.read_to_end(&mut buf)?;
+    from_slice(&buf)
+}
+
 /// Errors that can occur when deserializing a type.
-#[derive(Debug, PartialEq, Eq, Clone)]
+#[derive(Debug, Clone)]
```
alexcrichton, line 132: Unfortunately thtis removal is a breaking change, so we'll need to keep these traits.

src/de.rs
```diff
@@ -194,6 +240,9 @@ enum ErrorKind {
     /// Unquoted string was found when quoted one was expected
     UnquotedString,
 
+    /// An IO error occurred.
+    Io,
```
alexcrichton, line 244: I would prefer to not add a new error variant here because toml parsing in general doesn't have I/O errors, that's a higher-level concern. I'd personally prefer to add a new error type which is basically either an I/O error or a toml parsing error.

WhyNotHugo:
Updated. I'm not sure I'm convinced of the multiple layers of errors being returned though, I think adding the `Io` entry was simpler to deal with.

src/ser.rs
```diff
@@ -93,6 +94,60 @@ where
...
+///     toml::to_writer(&mut file, &config).unwrap();
+/// }
+/// ```
+pub fn to_writer<W, T: ?Sized>(writer: &mut W, value: &T) -> Result<(), Error>
+where
+    W: io::Write,
+    T: ser::Serialize,
+{
+    let b = to_vec(value)?;
+    writer.write_all(&b).unwrap();
```
alexcrichton, line 147: I don't think this will want to `.unwrap()`, I think it will want to propagate the error.

src/de.rs
```diff
@@ -98,6 +144,21 @@ struct ErrorInner {
...
 }
 
+/// Error that can occur when deserializing via IO.
+#[derive(Debug)]
+pub enum ReaderError{
+    /// An input/output error.
+    IoError(io::Error),
+    /// An internal serialisation error.
+    ParserError(Error),
+}
```
alexcrichton, line 154: I think that this will want an implementation of `std::error::Error` as well.

epage:

PR #439, opened 2021-08-15
Make datetime.rs structs and fields public #431

xpe:
In issue #431, @samuelcolvin wrote:

> I'd like to be able to access the fields of Datetime (and then obviously Date and Time) directly rather than having to re-parse the result of to_string().
>
> I don't need chrono integration, just direct access to the integer values.
>
> Would you accept a PR for this?

Then @alexcrichton replied:

> Seems reasonable to me to add!

Issue #431, opened 2021-06-07
make `Datetime` fields publically accessible


samuelcolvin:
I mentioned this on #159, but it's slightly different and I thought deserved a separate issue.

I'd like to be able to access the fields of `Datetime` (and then obviously `Date` and `Time`) directly rather than having to re-parse the result of `to_string()`.

I don't need chrono integration, just direct access to the integer values.

Currently in rtoml I'm using `std::mem::transmute` into duplicated structs in order to access the fields which as well as being ugly and `unsafe` means multiple copies.

For anyone else needing this, you can see the code [here](https://github.com/samuelcolvin/rtoml/blob/78ebad15f8d94b42fc7ada2a7ac7fd67e620216f/src/datetime.rs#L8-L38).

Would you accept a PR for this?

---

Also, (perhaps separate issue) I think `offset` should be moved into `date` since AFAIK times with timezone are not valid in toml.

alexcrichton:
Seems reasonable to me to add!

xpe:
@samuelcolvin I created a pull request over at #439

xpe:
@samuelcolvin This got merged in [83d787c60](https://github.com/alexcrichton/toml-rs/commit/83d787c60dbaaaf5b769f0948a5e8760ea5695cf). Does the PR meet your needs? If so, could you close this issue?

samuelcolvin:
It looks great, either your can close it now or I'll try and use it in rtoml next week.

xpe:
@samuelcolvin Good :). I can't close this, because I don't have "close issue" privileges. Sometimes I like to do some issue gardening.

ehuss:
Thanks @xpe!  It sounds like this is done, so I'll close it out.

not-my-profile:
Thanks for implementing this! I'd love to use it in a project I want to publish on crates.io, would it be possible to release a new version of the `toml` crate?

PR #455, opened 2022-01-26
Public datetime structs followups

not-my-profile:
Btw I'd love to use #439 in a project I want to publish on crates.io, would it be possible to release a new version of the `toml` crate?

### DIFF
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -120,7 +120,7 @@ pub const NAME: &str = "$__toml_private_Datetime";
 /// > ```
 ///
 /// [Local Date]: https://toml.io/en/v1.0.0#local-date
-#[derive(PartialEq, Clone)]
+#[derive(PartialEq, PartialOrd, Clone)]
 pub struct Date {
     /// Year: four digits
     pub year: u16,
@@ -150,7 +150,7 @@ pub struct Date {
 /// > must be truncated, not rounded.
 ///
 /// [Local Time]: https://toml.io/en/v1.0.0#local-time
-#[derive(PartialEq, Clone)]
+#[derive(PartialEq, PartialOrd, Clone)]
 pub struct Time {
     /// Hour: 0 to 23
     pub hour: u8,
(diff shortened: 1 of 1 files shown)


ehuss:
Can you say why this is adding `PartialOrd`?  That seems precarious, as there can be ambiguities or other confusions. For example, the ordering of a `Datetime` with just a date versus one with just a time. Is there precedent in other libraries for this?  `Value` itself is not sortable, so it is not clear why this is being added.

not-my-profile:
I did not add `PartialOrd` for `Datetime` to avoid that very ambiguity. I only added it for `Date` and `Time` for which it is unambiguous.

ehuss:
I see, but can you explain what your use case is?  It is helpful when submitting a PR to explain what it is changing and importantly why it is changing.

not-my-profile:
I am working on a static-site generator where each page can define metadata in TOML. That metadata can then be queried via a query language to e.g. sort pages by date. I am using `toml::Value` to store the metadata (since it can be arbitrary). I don't mind that `Value` is not sortable since I only want to allow comparisons between values of the same type anyway.

If `Date` and `Time` derive `PartialOrd` implementing that comparison becomes a bit more convenient for me.

Even better would be if `Datetime` was split as suggested in https://github.com/alexcrichton/toml-rs/issues/159#issuecomment-463732912 and all of these would implement `PartialOrd`. Should I attempt that instead?

not-my-profile:
To be fair I think that publicizing the `Datetime` struct as has been done in #431 is a bit weird because it allows the construction of `toml::Value`s that aren't valid TOML:

```rust
    let val = toml::Value::Table(
        vec![(
            "test".into(),
            toml::Value::Datetime(Datetime {
                date: None,
                time: None,
                offset: None,
            }),
        )]
        .into_iter()
        .collect(),
    );
    let out = toml::to_string(&val).unwrap();
    println!("{:?}", out);
```

prints `"test = \n"` ...

epage:

Issue #456, opened 2022-01-29
Docs for use with Serde incomplete


SeSodesa:
It would be nice if the Serde-related examples in the documentation worked out of the box. For example, simply inserting

	[dependencies]
	toml = "0.5"

into `Cargo.toml`, as instructed in the `toml-rs` [README](https://github.com/alexcrichton/toml-rs#readme), and then trying out the example

	use serde_derive::Deserialize;

	#[derive(Deserialize)]
	struct Config {
		ip: String,
		port: Option<u16>,
		keys: Keys,
	}

	#[derive(Deserialize)]
	struct Keys {
		github: String,
		travis: Option<String>,
	}

	fn main() {
		let config: Config = toml::from_str(r#"
			ip = '127.0.0.1'

			[keys]
			github = 'xxxxxxxxxxxxxxxxx'
			travis = 'yyyyyyyyyyyyyyyyy'
		"#).unwrap();

		assert_eq!(config.ip, "127.0.0.1");
		assert_eq!(config.port, None);
		assert_eq!(config.keys.github, "xxxxxxxxxxxxxxxxx");
		assert_eq!(config.keys.travis.as_ref().unwrap(), "yyyyyyyyyyyyyyyyy");
	}

given in <https://docs.rs/toml/latest/toml/#deserialization-and-serialization> produces the error

	λ cargo build
	   Compiling some_crate v0.1.0 (/path/to/crate)
	error[E0432]: unresolved import `serde_derive`
	  --> src/lib.rs:15:5
	   |
	15 | use serde_derive::Deserialize;
	   |     ^^^^^^^^^^^^ use of undeclared crate or module `serde_derive`

	For more information about this error, try `rustc --explain E0432`.
	error: could not compile `some_crate` due to previous error

In other words, the documentation should be more specific about which crates need to be included for the examples to work. What do I need to include into the dependencies-section for `toml-rs` to work with Serde?

KaniyaSimeji:
In `[dependencies]` of Cargo.toml
```serde = { version = "1.0", features = ["derive"] } ```
to the dependencies in Cargo.toml!

Air4x:
You should try to add `serde = {version="1", features=["derive"]}` in your Cargo.toml to be able to use `#[derive(Deserialize)]` in your code.

epage:
[...]

PR #457, opened 2022-01-30
Update docs for use with Serde

SeSodesa:
The required dependencies for use with Serde were not explicitly stated
before. This commit adds the lines needed in Cargo.toml for the Serde examples
to work, and modifies the use statements in the examples to reflect the
addition.

Hopefully solves #456.


ehuss:
Thanks, though this looks like it causes some errors.  Perhaps just keep it with `serde_derive` for now?

PR #464, opened 2022-05-24
speedup, test and improve datetime parsing

samuelcolvin:
I was looking for a datetime parsing library for pydantic-core and tested the datetime parsing logic from this library, I found it was already significantly faster than both chrono and iso8601, however I also discovered some fairly trivial changes could improve the performance significantly.

Changes proposed here:
* process `.bytes()` instead of `.chars()`, I haven't (yet) renamed the variable from `chars` to `bytes` to keep the changes to a minimum, I can if requested
* make `digit` a macro, this seemed to improve performance a little
* change the logic in nanosecond calculation to adjust the base just once at the end
* calculate the correct number of days in each month correctly, accounting for leap years
* add tests for the `Datetime` struct created when parsing strings
* add benchmarks for parsing datetimes, dates and times

Combined these changes have the following effect on performance:

**Before:**

```
test date              ... bench:          13 ns/iter (+/- 2)
test datetimes_as_toml ... bench:       2,316 ns/iter (+/- 11)
test dt_custom_tz      ... bench:          39 ns/iter (+/- 2)
test dt_naive          ... bench:          26 ns/iter (+/- 1)
test dt_z              ... bench:          26 ns/iter (+/- 1)
test time              ... bench:          16 ns/iter (+/- 0)
```

**After:**
```
test date              ... bench:           6 ns/iter (+/- 0)
test datetimes_as_toml ... bench:       2,250 ns/iter (+/- 8)
test dt_custom_tz      ... bench:          14 ns/iter (+/- 0)
test dt_naive          ... bench:          11 ns/iter (+/- 0)
test dt_z              ... bench:          12 ns/iter (+/- 0)
test time              ... bench:           9 ns/iter (+/- 0)
```

### DIFF
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -233,6 +233,15 @@ impl fmt::Display for Offset {
     }
 }
 
+macro_rules! digit {
+    ($bytes:ident) => {
+        match $bytes.next() {
+            Some(c) if b'0' <= c && c <= b'9' => c as u8 - b'0',
+            _ => return Err(DatetimeParseError { _private: () }),
+        }
+    };
+}
+
 impl FromStr for Datetime {
     type Err = DatetimeParseError;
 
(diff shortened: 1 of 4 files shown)


samuelcolvin:
I also noticed that `DatetimeParseError { _private: () }` is quite slow compared to returning an enum member or empty tuple, but I didn't want to change that in case this logic is required?

epage:

PR #467, opened 2022-07-15
Add ser::to_vec_pretty function

oriontvv:
Hi, thanks for awesome crate!

I've noticed that we miss pretty veriant of vec serialization like we do with `to_string_pretty/to_string`. So this PR fixes that.


epage:

PR #473, opened 2022-09-01
Add support for enum variants with parameters, fixes dotted tables as well

axos88:
This PR allows us to serialize/deserialize the TheEnum::Tuple, TheEnum::NewType, TheEnum::Struct variants as well:

```
#[derive(Debug, Deserialize, PartialEq)]
struct Val {
    val: TheEnum,
}

#[derive(Debug, Deserialize, PartialEq)]
enum TheEnum {
    Plain,
    Tuple(i64, bool),
    NewType(String),
    Struct { value: i64 },
}
```

The change possibly contains a breaking change because the enum TheEnum::Tuple used to be deserialized to

```
Val { val: TheEnum::Tuple(42, true) }
```

```
val = [42, true]
```
with loss of the variant information, so not enough information to deserialize.

The new behaviour is

```
  [val]
  Tuple = [42, true]
```

Which can now be deserialized.

Please note the change in `from_dotted_table`:

```
TheEnum::NewType("value".to_string())
```

currently does not deserialize correctly (UnsupportedType), and it only deserializes correctly from ```{ NewType = "value"}```. Without the braces it throws a 
```
thread 'enum_newtype::from_dotted_table' panicked at 'called `Result::unwrap()` on an `Err` value: Error { inner: ErrorInner { kind: UnquotedString, line: Some(0), col: 0, at: Some(0), message: "", key: [] } }', tests/enum_external_deserialize.rs:143:52
```

Might need more work around serialization?

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -588,6 +588,10 @@ impl<'de, 'b> de::SeqAccess<'de> for MapVisitor<'de, 'b> {
             })
             .unwrap_or(self.max);
 
+        let cur = if self.tables[self.cur_parent].values.clone().unwrap_or(Vec::new()).is_empty() && self.cur_parent + 1 < self.max {
+            self.cur_parent + 1
+        } else { self.cur_parent };
+
         let ret = seed.deserialize(MapVisitor {
             values: self.tables[self.cur_parent]
                 .values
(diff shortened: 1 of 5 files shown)


axos88:
@alexcrichton , I'd appreciate some feedback on this, and also I'd need a little bit of help on how to `serialize_struct_variant` into a nested array.

alexcrichton:
Sorry I do not have the time/motivation to review and land larger changes like this. I try to perform maintenance (which albeit I'm not great at with this crate), but for major features like this I think a "new official" maintainer would need to step up.

axos88:
@alexcrichton I understand. Thanks for your work on this up to date. Although I also don't have the resources to take over this project, it would be great to mention that this is not actively maintained / major changes are not accepted / looking for a new maintainer in the readme.

epage:

PR #430, opened 2021-06-02
Integrate toml-rs with OSS-Fuzz

evverx:
It should make it possible to fuzz toml with `cargo fuzz` locally and OSS-Fuzz continuously. It's related to  https://github.com/google/oss-fuzz/pull/5865

PR #476, opened 2022-09-16
cargo-fuzz support

anfedotoff:
Hi!
I'm using your crate it's great! Maybe add some fuzzing support?


anfedotoff:
Found this PR #430. Fuzzing support is already opened.

